PR #36, opened 2018-09-22
Dev

CBernsCode:


jasondkiesling:
What is supposed to be at learnmyr.org/first-time? When I go to that URL or localhost:3000/first-time I get a 404.

Issue #124, opened 2019-03-28
Create classroom style view


jasondkiesling:
Create the ability to add a classroom code to a scene and allow the classroom code owner to view those scenes via a special view.

jasondkiesling:
Fixed in #128

PR #129, opened 2019-03-28
Classroom submissions

jasondkiesling:
Added the ability for users to create a classroom code of their choice and for users to add that code to their project. The owner of the classroom can then see all projects in one place.

Addresses: #124


CBernsCode:
It looks really great. 

I am a bit concerned about the selection process for the classroom, mainly as it pertains to being able to provide more details. It seems like the case where names may be reoccurring and they could be confusing. I think a drawer or a select list that supports a second line might be helpful.

jasondkiesling:
I completely agree. It was something I thought about. I didn't know what to put there however. Google doesn't allow you to look up a user by UID or else I would have figured out how to list their email with it.

In the future, I want to incorporate the ability to see the project description and the project share menu. As a "solution" for now, I forsee teachers having the students add their name or initials to the project name.

PR #162, opened 2019-05-21
Modified reference section to be more clear.

elenaizotova:


jasondkiesling:
I'm not sure why the tests failed on this. I need to do some more investigating as to why that happened. I will override that however for this.

Issue #140, opened 2019-04-02
MYR tour can't highlight editor when editor is closed


jasondkiesling:
**Steps To Reproduce:**
1. Open MYR
2. Close the editor
3. Launch the tour

jasondkiesling:
Per 5/29 discussion with @elenaizotova and @vrindle, it is fine to open the editor during the tour, but it must also close it afterwards. It would also be useful to include a note telling the user how to open/close it.

PR #163, opened 2019-05-21
fixed the bug where opening tour while editor is closed didn't highlight editor

samuel-zuk:
ignore the weird commit history, i'm just learning how to use git lol :+1:


jasondkiesling:
Also, can we remove the console.log in src/actions/sceneActions.js. That was my fault for leaving that in.

jasondkiesling:
Addresses #140

samuel-zuk:
honestly, i think I only changed those functions to arrow functions to try to learn the syntax. you don't need to keep that change.

jasondkiesling:
Can you revert that change and push it to GitHub? I can't update it on your fork. It's best to keep things all following the same syntax.

Issue #89, opened 2019-03-06
Changing grid/floor/scene options before rendering/saving code loses code in buffer


fgmart:

jasondkiesling:
This issue has been expanded to include almost all of the options within the scene config menu.

PR #172, opened 2019-05-23
Fixes tour issue with editor closed and lost changes with scene config menu updates

samuel-zuk:
should fix these bugs: [#140](https://github.com/engaging-computing/MYR/issues/140#issue-428307710) and [#89](https://github.com/engaging-computing/MYR/issues/89#issue-417986347).


src/components/MyrTour.js
```diff
@@ -21,6 +21,7 @@ class MyrTour extends Component {
           steps={steps}
           maskClassName="mask"
           isOpen={this.state.isTourOpen}
+          onAfterOpen={this.props.isEditorClosed ? this.props.toggleEditor : ()=>{}}
```
jasondkiesling: It looks like there is onAfterClose as well. Can you add that and have it toggle the editor closed after it opens it? It would also be good to not in the tour content how the user can do that themselves.

src/components/Header.js
```diff
@@ -651,14 +651,17 @@ class Header extends Component {
               <Icon className="material-icons">perm_media</Icon>
             </IconButton>
           </Tooltip>
-          <MyrTour />
+          <MyrTour 
+            isEditorClosed={this.props.scene.settings.viewOnly}
```
jasondkiesling: For consistency sake, while you're making the other edit can isEditorClosed be renamed to viewOnly? That way it's one term in all places.

src/components/SceneConfigMenu.js
```diff
@@ -263,7 +263,10 @@ class ConfigModal extends Component {
     return (
       <ButtonBase
         style={style}
-        onClick={() => this.props.sceneActions.toggleCoordSky()} >
+        onClick={() => {
+          this.props.handleRender();
```
jasondkiesling: Can we include this logic in all of the menu options? Sky color, floor color, hiding the editor, and removing the floor all also have the same issue as #89.

jasondkiesling:
These have been confirmed as working as described on Ubuntu 18.04 LTS with the Chromium browser.

src/components/MyrTour.js
```diff
@@ -20,8 +25,13 @@ class MyrTour extends Component {
         <Tour
           steps={steps}
           maskClassName="mask"
-          isOpen={this.state.isTourOpen}
-          onAfterOpen={this.props.isEditorClosed ? this.props.toggleEditor : ()=>{}}
+          isOpen={this.state.isTourOpen}  
+          onAfterOpen={()=>{
+            this.viewOnlyOnOpen = this.props.viewOnly;
```
jasondkiesling: What are you trying to do with this line here? If you're trying to set the state, you need to use `this.setState({viewOnlyOnOpen: this.props.viewOnly})`. However, you did already set that in the constructor.

src/components/MyrTour.js
```diff
@@ -39,7 +49,11 @@ const steps = [
   {
     selector: '#ace-editor',
     content: 'This is the editor. You can create 3D scenes using JavaScript ' +
-      'and a special set of instructions or functions to MYR.',
+      'and a special set of instructions or functions to MYR.'
+  },
+  {
+    selector: '#configure-scene',
```
jasondkiesling: This could end up a little confusing where we go to the scene config menu button twice. Let's just add this note within the editor tour itself.

src/components/SceneConfigMenu.js
samuel-zuk: my last commit added this logic for changes to sky/floor color and removing the floor, but not for hiding the editor, as all of the other settings require the scene to be re-rendered anyway. i don't believe hiding the editor does, and i believe @jSwanson99's fix to toggling the editor not saving works better for this specific case.

src/components/MyrTour.js
samuel-zuk: fair enough. will do.

src/components/MyrTour.js
samuel-zuk: i realized the tour object was created when the page loads, rather than when the tour button is clicked. since the editor is always open by default when the page loads, viewOnlyOnOpen was always evaluating to false b/c its value was only set when the tour object was created. this line checks if the editor is open or not and updates viewOnlyOnOpen to match. it would probably be better off in the component state anyway though, so i will move it there.

PR #175, opened 2019-06-03
Fixes bug where save buffer is not updated when entering view only

jSwanson99:


src/components/SceneConfigMenu.js
```diff
@@ -222,14 +222,18 @@ class ConfigModal extends Component {
 
   // Toggles whether the editor is showing
   viewToggle = () => {
-    let style = this.props.scene.settings.viewOnly ? btnStyle.off : btnStyle.on;
+    let viewOnly = this.props.scene.settings.viewOnly
```
jasondkiesling: This is not technically wrong, but it adds unnecessary complexity to the code. It is best to only use a variable to store a prop when you're using it multiple times or in long strings. In addition, props stored as a variable should always be a const as you can't change them.

src/components/View.js
```diff
@@ -171,33 +170,35 @@ class View extends Component {
 
   render = () => {
     return (
-      <a-scene physics="debug: false; friction: 3; restitution: .3;" embedded debug="false">
```
jasondkiesling: What's up with these changes?

src/components/layouts/Ide.js
```diff
@@ -6,46 +6,60 @@ import View from '../View';
...
+                    projects={this.props.projects}
+                    courses={this.props.courses}
+                    classroomActions={this.props.classroomActions}
+                    classrooms={this.props.classrooms}
+                    layoutType={layoutTypes.IDE}
+                />
+
+                <div className="row no-gutters">
+                    {
+                        /*
```
jasondkiesling: Let's move comments about issues to Github. Code comments are best when they more about what the code does or quick issues. Feel free to make a reference to Github by citing the issue number.

src/components/layouts/Ide.js
```diff
@@ -6,46 +6,60 @@ import View from '../View';
 
 import * as layoutTypes from '../../constants/LayoutTypes.js';
 
-export const Ide = ({ editor, editorActions, user, authActions, scene, sceneActions, projectActions, courseActions, projects, courses, match, classroomActions, classrooms }) => (
```
jasondkiesling: There are also a lot of format changes here. What are these for?

src/components/Editor.js
```diff
@@ -10,6 +10,12 @@ import customCompleter from './customCompleter.js'
 *
 */
 class Editor extends Component {
+  componentWillUnmount() {
+    //Text is saved in state when editor unmounts / when in view only.
+    //This does not fully 'save' it because state will be lost when we
```
jasondkiesling: Although this is a valid, this is something that's I'm working on fixing. This comment can be simplified to say that it updates the reducer state before closing the editor.

src/components/View.js
jSwanson99: These are just whitespace changes I didn't intentionally make, is there a way to ignore/revert these easily?

PR #184, opened 2019-06-04
Moved animation properties to the cursor and added shiftToColor()

elenaizotova:


src/myr/reference.js
```diff
@@ -168,57 +168,126 @@ let transformations = [
...
+    description: <span>Updates the magnitude attribute in the cursor, in degrees and general units. Default: 1</span>
+  },
+  {
+    name: <span>spin(<span className="string">element</span>)</span>,
     description: <span>Spins the element around the y axis <span className="number">magnitude</span> degrees.</span>
   },
   {
-    name: <span>yoyo(<span className="string">element</span>, <span className="number">magnitude</span>, <span className="bool">loop</span>, <span className="number">duration</span>)</span>,
+    name: <span>yoyo(<span className="string">element</span>)</span>,
     description: <span>Bounces the element <span className="number">magnitude</span> units in a positive direction on the y axis for <span className="number">duration</span> milliseconds.</span>
```
jasondkiesling: These descriptions should be updated to reflect that the values are no longer in the function but from the cursor.

Issue #183, opened 2019-06-04
Tour needs to close reference if open


jasondkiesling:

jasondkiesling:
@samuel-zuk This is similar to your work in #172. When the tour opens, it can just close the reference automatically. One way to do this is to create a function to set the state of the reference isOpen to false and call that on line 29 in src/components/MyrTour.js close the reference with the on click.

PR #191, opened 2019-06-05
Fixed bug where tour could open with reference open, and also disables tour when reference page was opened.

samuel-zuk:
addresses bug [#183](https://github.com/engaging-computing/MYR/issues/183).

### DIFF
--- a/src/components/Reference.js
+++ b/src/components/Reference.js
@@ -22,13 +22,10 @@ export default class Reference extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      open: false,
       value: 'a',
     };
     this.tableData = myrReference();
   }
-
-  handleToggle = () => this.setState({ open: !this.state.open, value: 'a' });
 
   handleChange = (event, value) => {
     this.setState({ value });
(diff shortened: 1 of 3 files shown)


src/components/Reference.js
```diff
@@ -118,7 +115,10 @@ export default class Reference extends React.Component {
               icon={<Icon className="material-icons">close</Icon>}
               label="CLOSE"
               value='x'
-              onClick={this.handleToggle} />
+              onClick={() => {
+                this.props.handleReferenceToggle();
+                this.setState({ value: 'a' });
```
jasondkiesling: What is this line doing?

jasondkiesling: (Line 120)

jasondkiesling: The question is less so as to what it is doing but more so as why is it doing it.

samuel-zuk: if this line wasn't included, closing the reference using the close button and then re-opening the reference would bring up [this page](https://imgur.com/34i0F7v), which shouldn't happen because there is no content on the non-existent "close" tab. idk why this happens but setting the reference page back to page "a" (i.e. the geometry page) when the close button is clicked fixes this.

samuel-zuk: i could try to figure out why this only happens now, after i lifted the state of the reference component up to the header, but seeing as everything else works as expected besides this one weird bug, this was a super quick and easy fix that clears that up.

jasondkiesling: Ah. I know why that's happening. It's because close says to open tab "x" which is non-existent.

PR #192, opened 2019-06-05
Made close and open in new tab buttons in reference colored

elenaizotova:


jasondkiesling:
Fixes #193

src/components/SceneConfigMenu.js
```diff
@@ -398,6 +399,7 @@ class ConfigModal extends Component {
       <div>
         {!isDisabled ?
         <div>
+          <Tooltip title = "Scene Config Menu">
```
jasondkiesling: Although this is what we refer to it as internally, it is referred to officially as either "Scene Settings" or "Scene Configuration".

PR #197, opened 2019-06-06
Changes can be saved in view only & save buffer not lost switching view mode & crtl f enabled in editor

jSwanson99:
also resolved merge conflicts


src/reducers/index.js
```diff
@@ -18,9 +18,11 @@ const reducer = combineReducers({
   classrooms
 });
 
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ trace: true, traceLimit: 25 }) || compose;
```
jasondkiesling: What is this actually doing?

yarn.lock
```diff
@@ -2303,6 +2303,11 @@ caseless@~0.12.0:
   version "0.12.0"
   resolved "https://registry.yarnpkg.com/caseless/-/caseless-0.12.0.tgz#1b681c21ff84033c826543090689420d187151dc"
 
+chain-function@^1.0.0:
```
jasondkiesling: What were all of these added for? There are a lot of packages added.

src/components/Header.js
```diff
@@ -645,7 +656,7 @@ class Header extends Component {
           <Tooltip title="Open" placement="bottom-start">
             <IconButton
               id="open-btn"
-              onClick={this.handleLoadToggle}
+              onClick={this.handleLoadTfoggle}
```
jasondkiesling: We're back with this issue again.

src/components/View.js
```diff
@@ -171,33 +170,35 @@ class View extends Component {
...
-            <a-cylinder checkpoint radius="1" height="0.3" position="-25 1 -25" color="#39BB82"></a-cylinder>
-            <a-cylinder checkpoint radius="1" height="0.3" position="25 1 25" color="#39BB82"></a-cylinder>
-            <a-cylinder checkpoint radius="1" height="0.3" position="-25 1 25" color="#39BB82"></a-cylinder>
-            <a-cylinder checkpoint radius="1" height="0.3" position="25 1 -25" color="#39BB82"></a-cylinder>
-            <a-circle checkpoint radius="1" rotation="90 0 0" position="0 10 0" color="#39BB82"></a-circle>
-          </a-entity>
-          : null
-        }
-      </a-scene>
+        <a-scene physics="debug: false; friction: 3; restitution: .3;" embedded debug="false">
```
jasondkiesling: Can you run the auto formatter in VSCode and make sure this is how the formatting should be? It could have been wrong this entire time, but I'm not sure why there is such a whitespace difference.

package.json
```diff
@@ -29,6 +29,7 @@
     "qrcode.react": "^0.8.0",
     "react": "^16.8.4",
     "react-ace": "^5.9.0",
+    "react-beforeunload": "^1.0.4",
```
jasondkiesling: Do we actually need this package?

src/reducers/index.js
jSwanson99: its for the redux devtools, so you can see the state changes as they happen.

yarn.lock
jSwanson99: I think I had added some packages and then not used them. they should be removed now

src/components/Header.js
jSwanson99: fixed, sorry

src/components/View.js
jSwanson99: done

package.json
jSwanson99: no sorry, i just removed it

src/reducers/index.js
jasondkiesling: Interesting. Good to know.

src/components/Editor.js
```diff
@@ -5,11 +5,18 @@ import 'brace/theme/github';
...
+
 /**
 * @summary - Editor is a React Component that creat the Ace Editor in the DOM.
 *
 */
+
 class Editor extends Component {
+  componentWillUnmount() {
+    //Updates state in reducer before closing editor
+    this.props.refresh(window.ace.edit("ace-editor").getSession().getValue(), this.props.user ? this.props.user.uid : 'anon');
```
jasondkiesling: I'm testing this now. It does do what it says. However it is a little confusing as it doesn't also render. Can you add a render cycle here too?

jSwanson99: Should be done now

PR #221, opened 2019-06-10
Reference examples now open in new tabs

elenaizotova:


src/components/reference/Reference.js
```diff
@@ -41,7 +41,8 @@ export default class Reference extends React.Component {
       let link = '/reference/' + example;
       return (
         <IconButton 
-        href={link} 
+        href={link}
+        target="_blank"
```
jasondkiesling: Easy fixes are the best fixes.

Issue #71, opened 2019-01-16
Saving on courses


jasondkiesling:
Saving on courses should auto-populate a name and not open the save as menu

jasondkiesling:
This should be easier to do now with the modified scene reducer

PR #224, opened 2019-06-12
Saving on course will auto populate the title of the Lesson on saving tab #71

Jlu18:


Jlu18:
My latest commit only shows the changes I made.

jasondkiesling:
I'll try to look at this tonight. I need to check something.

jasondkiesling:
Also, in the future please try to avoid using issue numbers in commit messages. If trying to see the reason for a change, it requires leaving the IDE to go to GitHub to look up the issue number.

src/components/structural/header/Header.js
```diff
@@ -1,707 +1,707 @@
-import React, { Component, Fragment } from 'react';
```
jasondkiesling: This shouldn't have had any changes. Can you see what's up with this diff?

jasondkiesling: I'm suspecting that this might be able to be fixed by running the VSCode autoformatter. That often fixes this type of issue.

Jlu18: I run autoformatter in VSCode and I don't see any changes in Header.js file.

jasondkiesling: Did you delete the header file at all and then re-add it?

This is part of the output of the diff.
```--- a/src/components/structural/header/Header.js
+++ b/src/components/structural/header/Header.js```

Jlu18: No, I didn't delete the all Header.js file but the part where I add the code in the first commit.

jasondkiesling: You might need to reset and recreate the commit. I'm not sure what's going on with that.

Jlu18: I made a new branch with the exact same change and it doesn't highlight anything from header.js. Can I create a new pull request base of that branch?

jasondkiesling: Yes

PR #229, opened 2019-06-13
Fixed makeUnPushable/unDroppable bug

jSwanson99:
… and added a loading screen/overlay while waiting for view to mount

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "aframe-react": "^4.4.0",
     "bootstrap": "^4.0.0",
     "cannon": "^0.6.2",
-    "create-react-app": "^1.5.2",
+    "create-react-app": "^3.0.1",
     "firebase": "^5.5.7",
     "jquery": "^3.3.1",
     "qrcode": "^1.2.0",
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -324,6 +324,16 @@ class Myr {
     return outerElId;
   }
 
+  // Disallows the entity to be pushed
+  makeUnPushable = (outerElId) => {
+    let el = this.getEl(outerElId);
+    if (el["force-pushable"]) {
+      el["dynamic-body"] = null;
+      el["force-pushable"] = "false";
+    }
+    return outerElId;
+  }
+
   push = (outerElId, x, y, z) => {
     // Add an event listener
     document.addEventListener('myr-view-rendered', (e) => {
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -74,7 +74,6 @@ class Myr {
   * @summary - Reset this.els to the base elements supplied to the constructor
   */
   reset = () => {
-
     // Reset base params, we might be able to merge two objects later
     this.id = 0;
     this.color = 'red';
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "aframe-react": "^4.4.0",
     "bootstrap": "^4.0.0",
     "cannon": "^0.6.2",
-    "create-react-app": "^1.5.2",
+    "create-react-app": "^3.0.1",
```
jasondkiesling: Was there a particular reason for updating this?

src/components/Header.js
```diff
@@ -30,6 +30,8 @@ const exitBtnStyle = {
   right: 0,
 };
 
+let shouldRender = false;
```
jasondkiesling: This should be part of the component state. Then whenever you need to change it you should use this.setState({}). Having anything outside of the class that isn't a const can cause issues.

src/components/Header.js
```diff
@@ -180,6 +182,10 @@ class Header extends Component {
     if (this.state.lastMsgTime !== this.props.message.time && this.props.message.text !== "") {
       this.setState({ snackOpen: true, lastMsgTime: this.props.message.time });
     }
+    if(shouldRender) {
```
jasondkiesling: This doesn't look like it is really the way that we should be doing this. Something is weird.

src/components/View.js
```diff
@@ -12,15 +12,24 @@ import * as THREE from 'three';
  * system utilizes the entity compoent system(ECS) to build objects in the scene from different
  * components.
  */
+
+ let isLoading = true;
```
jasondkiesling: This also should be part of the component state.

src/components/View.js
```diff
@@ -169,7 +178,7 @@ class View extends Component {
   }
 
   render = () => {
-    return (
+    const aScene = (
```
jasondkiesling: Is this actually a const?

src/components/View.js
```diff
@@ -200,6 +208,21 @@ class View extends Component {
...
       </a-scene>
     );
+  
+    const spinner = 
+      <span className='spinner'>
+        <div className='cube1'></div>
+        <div className='cube2'></div>
+      </span>
+
+    let ret = null;
```
jasondkiesling: This can be reduced to `return isLoading ? spinner : aScene;`.

src/components/View.js
jasondkiesling: This variable name is also unclear. Let's make it more clear.

src/components/layouts/Ide.js
```diff
@@ -31,9 +31,9 @@ export const Ide = ({ editor, editorActions, user, authActions, scene, sceneActi
             {
                 scene.settings.viewOnly
                     ?
-                    <div id="scene" className="col-12" >
-                        <View objects={editor.objects} sceneConfig={scene} assets={editor.assets} />
-                    </div>
+                        <div id="scene" className="col-12" >
```
jasondkiesling: Was this run through the formatter?

src/myr/Myr.js
```diff
@@ -289,17 +289,39 @@ class Myr {
   makeDroppable = (outerElId, mass = 2) => {
     let el = this.getEl(outerElId);
     let dynamicBody = `shape: auto; mass: ${mass}; angularDamping: 0.5; linearDamping: 0.5;`;
-    el["dynamic-body"] = dynamicBody;
+    el["dynamic-body"] = dynamicBody;
```
jasondkiesling: Please remove unneeded whitespace.

src/myr/Myr.js
```diff
@@ -289,17 +289,39 @@ class Myr {
...
     let el = this.getEl(outerElId);
-    el["dynamic-body"] = "";
+    //Only makes an item undroppable if it is droppable but is not pushable
+    if(el["dynamic-body"] && (!el["force-pushable"] || el["force-pushable"] === "false")) {
+      el["dynamic-body"] = null;
+    }
     return outerElId;
   }
 
+  // Allows the entity to be pushed
```
jasondkiesling: Was this run through the formatter? There is an option to auto-format on save that is best if it is turned on.

yarn.lock
```diff
@@ -2314,7 +2314,15 @@ chalk@2.4.1:
     escape-string-regexp "^1.0.5"
     supports-color "^5.3.0"
 
-chalk@^1.1.1, chalk@^1.1.3:
```
jasondkiesling: What's the reason for updating all of these?

package.json
jSwanson99: I was having problems with create react app after installing eslint, so I had no remove it then install create-react-app again. I can try and go to an older version if needed

src/components/View.js
jSwanson99: I'm not sure what you mean

yarn.lock
jSwanson99: They might have been updated when I reinstalled create-react-app, I didn't intentionally update them.

jSwanson99:
I removed the loading screen because I couldn't get it to work through the components state, so I want to take more time to figure out a fix for that. Also any white space that was added/remove in this commit was due to the auto formatter

jasondkiesling:
The way that render is working now isn't really the way we should be doing it. More investigation is needed. If you want to remove that part for now we can push the makeUnDrop/Pushable part.

jasondkiesling:
You also have some merge conflicts that need to be resolved.

PR #234, opened 2019-06-15
Project drawer moved to a tabbed modal

elenaizotova:


src/components/reference/Reference.js
```diff
@@ -85,87 +85,87 @@ export default class Reference extends React.Component {
...
-              style={{ background: "red", color: "white" }}
-              icon={<Icon className="material-icons">close</Icon>}
-              label="CLOSE"
-              value='x'
-              onClick={() => {
-                this.props.handleReferenceToggle();
-                this.setState({ value: 'a' });
-              }}/>
-          </Tabs>
+        {!isDisabled ?
```
CBernsCode: When doing a really long ternary like this I usually like to put the "?" and ":" on a new line. You can also use "&&" after the cond, rather than using null. It's a style choice but wanted to give some feedback.

PR #238, opened 2019-06-17
Fixed bug where view only toggle doesn't work in courses

jSwanson99:


jSwanson99:
I'm not sure what the two merges are about.. they happened when I tried to git pull and it gave me the option to cancel, so I did because I didn't intend to merge anything.

jasondkiesling:
Merges are because you pulled new changes from GitHub into your code. Cancelling means you cancelled changing the default merge commit message.

src/components/layouts/Guided.js
```diff
@@ -10,6 +10,7 @@ import * as layoutTypes from '../../constants/LayoutTypes.js';
 export const Guided = ({ editor, user, scene, editorActions, authActions, projectActions, projects, courseActions, courses, course, match, sceneActions, classroomActions, classrooms }) => (
     <div className="App">
         <Header
+            viewOnly={scene.settings.viewOnly}
```
jasondkiesling: Why does the header need this?

jSwanson99: For the handleSave function, I previously made it so it would check to see if its in view only so it knows to get the text from the editor or from the state. This fixed the bug where you couldn't save in view only

PR #241, opened 2019-06-18
Implements a loading screen overlay

jSwanson99:
This is slightly different than the loading screen implementation I had tried before, I moved it to the routes.js file because then it will trigger when anything loads (including anything we may add later on). To do this I had to make the routes.js file a class component, because it needs the lifecycle functions. If we don't want to clutter that file I could revert the changes & put them in View, but then it will only apply to anything that tries to load the View.


src/routes.js
```diff
@@ -6,19 +6,55 @@ import Classroom from './containers/Classroom';
...
-        <Route path='/class/:classroom' component={Classroom} />
-        <Route path='/reference/:function' component={ReferenceExample} />
-        <Route path='/:id' component={Ide} />
-        <Redirect from="/lesson" exact to="/" />
-        <Redirect from="/course" exact to="/" />
-      </Switch>
-    </BrowserRouter>
-  );
-};
+class Routes extends React.Component {
```
jasondkiesling: I think it's best if we leave the router as just a router. I would like the logic for this to be within just the view component itself.

jSwanson99: got it, should be done

jSwanson99:
Closing this because I just realized this (loading screen) unexpectedly broke some things

PR #248, opened 2019-06-19
Adds functionality for subtractive geometries

samuel-zuk:
addresses [#174](https://github.com/engaging-computing/MYR/issues/174)


src/components/editor/customCompleter.js
```diff
@@ -79,6 +79,12 @@ export const customCompleter = {
             "group()",
             "getRandomColor()",
             "setPhiLength()",
+            "makeDroppable()",
+            //"makeUnDroppable()",
```
jasondkiesling: These should be able to stay in.

src/myr/reference.js
```diff
@@ -205,6 +205,14 @@ let transformations = [
   //  description: <span>Prevents the <span className="string">element</span> from being pushed by the cursor, but allows the element to be affected by physics.</span>
   //},
   //Not quite sure about the inner workings of this one yet
+  {
+    name: <span>makeSubtractive(<span className="string">element</span>)</span>,
+    description: <span>Makes the <span className="string">element</span> erase other shapes that it overlaps with, instead of being drawn as a new shape.</span>
```
jasondkiesling: Is there a better word than erase?

samuel-zuk: i was going to put "subtract from" but i wasn't sure how clear that would be.

src/myr/Myr.js
```diff
@@ -314,14 +314,22 @@ class Myr {
     return outerElId;
   }
 
-  // Disallows the entity to be pushed
-  makeUnPushable = (outerElId) => {
```
jasondkiesling: Why are we taking out makeUnPushable?

src/components/structural/header/MyrTour.js
```diff
@@ -94,7 +94,8 @@ const steps = [
   },
   {
     selector: '#configure-scene',
-    content: 'Modify and share your scene including setting the background color, enabling the grid, and enabling flying.',
+    content: 'Modify and share your scene including setting the background color, ' +
+      'enabling the grid, and enabling flying.',
```
jasondkiesling: Was there a specific reason for this change?

src/myr/Myr.js
```diff
@@ -590,7 +637,8 @@ class Myr {
       position: this.position,
       scale: this.scale,
       rotation: this.rotation,
-      material: `color: ${this.color};  side: double;`,
+      material: `color: ${this.color};`,
```
jasondkiesling: Why did we take out `side:double`?

src/myr/Myr.js
```diff
@@ -452,7 +489,8 @@ class Myr {
...
```
jasondkiesling: Same here with `side:double`.

src/myr/reference.js
jasondkiesling: Let's make it subtract from. It's not really erasing. We made the reference examples for if someone is confused.

src/components/editor/customCompleter.js
jasondkiesling: Yes.

src/myr/Myr.js
samuel-zuk: that was an accident which should be fixed now

src/myr/Myr.js
samuel-zuk: i had figured out that i put that there early on when trying to figure out a-frame's entity-component system and looking back it serves no useful purpose. it made the tests fail so i took it out.

src/myr/reference.js
samuel-zuk: will do!

package.json
```diff
@@ -1,64 +1,65 @@
 {
-    "name": "myrproject",
```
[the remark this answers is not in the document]
samuel-zuk: i'm not 100% sure why it's like that other than that i ran "yarn lint --fix"

src/myr/Myr.js
jasondkiesling: What side:double does it makes the element have color on both the inside and outside. To see this you can go inside a shape and see if it has color or not.

samuel-zuk:
>     * Let's leave the side as double for all of the objects. It's something that we have also been talking about adding before.

gotcha. i'll update the tests to reflect this change

package.json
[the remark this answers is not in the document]
jasondkiesling: The formatting here still needs to be fixed. It doesn't run eslint on this file yet.

src/components/classroom/Classroom.js
```diff
@@ -137,7 +137,7 @@ class ClassroomModal extends Component {
 
     deleteClassroom = () => {
         const userClassrooms = this.props.classrooms.classrooms;
-        let optionItems = [];
+        const optionItems = [];
```
jasondkiesling: Some of these might not actually be const. This could cause issues. I know eslint gave a warning but it's something to be careful when using --fix.

src/components/structural/header/Header.js
```diff
@@ -172,583 +172,583 @@ class Header extends Component {
...
-        );
-    }
-
-    //You can pass functions into this in order to have 
-    //multiple setState/state actions dispatched within an event handler
-    //Currently only used for render button
-    postpone(f) {
-        window.setTimeout(f, 0);
-    }
+  /**
```
jasondkiesling: This still needs to be fixed. Make sure it's 4 spaces per indent.

src/myr/Myr.js
```diff
@@ -44,780 +45,890 @@ class Myr {
...
+              // Collision free so we can bind to window
+              window[element] = this[element];
+          }
+      });
+  }
+
+  /**
+  * @summary - Reset this.els to the base elements supplied to the constructor
+  */
+  reset = () => {
```
jasondkiesling: Same spacing issue here.

jasondkiesling:
As this is still a work in progress, I'm going to close this for now. Feel free to open it again as a draft pull request if you need help or a regular pull request when you're ready for testing.

PR #252, opened 2019-06-20
Editor with change will now show the confirmation window before exit

Jlu18:
This confirmation dialog will show up any action that will refresh the pages if the editor is not saved.


Jlu18:
I purposely didn't remove the confirm window on the course because going to next lesson doesn't refresh the page which doesn't take effect of the EventListener. So it won't duplicate the check but the confirmation will show every time user try to go to next or previous lesson with or without any change (which may be irritating for some user).

PR #254, opened 2019-06-20
project with same name is now sort by its timestamp to be consistent

Jlu18:


jasondkiesling:
We have some old scenes that don't have .data.ts. I'll do some checking to make sure that we don't break because of that.

jasondkiesling:
The logic is definitely right. We might just have to do some checking before running that check.

PR #258, opened 2019-06-21
Implement relative cursor modification functions

Jlu18:
These files are sample code for example scene in the reference page.

[increasePosition.txt](https://github.com/engaging-computing/MYR/files/3314521/increasePosition.txt)
[increaseXPos.txt](https://github.com/engaging-computing/MYR/files/3314522/increaseXPos.txt)
[increaseYPos.txt](https://github.com/engaging-computing/MYR/files/3314523/increaseYPos.txt)
[increaseZPos.txt](https://github.com/engaging-computing/MYR/files/3314524/increaseZPos.txt)


src/myr/Myr.js
```diff
@@ -153,6 +153,53 @@ class Myr {
         }
     };
 
+    increasePosition = (x = 0, y = 0, z = 0) => {
+
+        if (typeof x === "number" && typeof y === "number" && typeof z === "number") {
```
jasondkiesling: In case the user also wants to use the new position for another purpose, let's also return the new position for each of these.

src/myr/reference.js
```diff
@@ -135,6 +135,28 @@ let transformations = [
         description: <span>Changes the position of the cursor along the Z axis to the given value.</span>,
         example: "setZPos"
     },
+    {
+        name: <span> increasePosition(<span className="number">x</span>,<span className="number">y</span>,<span className="number">z</span>)</span>,
+        description: <span>Changes the position of the cursor relative to the current position. This function takes both positive and negative values. </span>,
+        example: "increasePosition"
```
jasondkiesling: Can you remove the example line? I'll add that myself once the examples are in the database.

src/myr/Myr.js
jasondkiesling: (So the position for increasePosition and the x, y, or z coordinate for the rest of them.

Jlu18: Should the return type for the increasePosition() be an array or object? I think array is better since you can use the array as an argument using the spread operator but what you think? 

Ex. 
`let x = increasePosition(2,2,2);`
`setPosition(...x); //same position`

jasondkiesling: Let's return it as an object. If they're setting the position twice, that's on them. There isn't really a reason to do that. I was thinking more for animations where as you go up in the y axis you might increase your width. There is a scene that does this but I can't think of it. A user might want to use the new y coordinate to specify the width of the next entity.

PR #262, opened 2019-06-22
1.3.0 | New relative cursor modification functions, editor bug fix, welcome screen, project view updates, reference UI update

jasondkiesling:


samuel-zuk:
Tested at 12:15 AM on (a derivative of) Ubuntu Linux on Firefox:
* all the buttons work on the welcome screen
* timestamps on the cookies are correct for each option
* the example dropsies scene within the popup doesn't have functional wasd/arrow controls for movement unless you enter VR 
* tabbed project modal and redesigned reference page work as expected
* i wasn't able to break the new increasePosition functions, so they should be fully functional

samuel-zuk:
Tested on Chrome and Firefox on Windows 10 at 10:40 and got the same results as above

jSwanson99:
Tested on Opera on Windows 10:
- Welcome screen works as expected (tested this @ 10 am)
- increasePosition & other increase functions, reference, and project modal all work as expected

- Was not able to replicate the wasd/arrow key buy in dropsies example scene

jSwanson99:
Correction: I was able to replicate the wasd bug in the dropsies scene, within the welcome screen

samuel-zuk:
Tested again on Firefox for Linux at 1:00 PM, and the example scene within the welcome splash now isn't letting me look around either (it did before). Not sure what caused this.

samuel-zuk:
Tested on Chrome for Android (Pie) a few times between 1:00 and 1:10, all buttons work as expected, the example scene in the welcome splash is slow to load but after a little while, you can click on it and look around. I was able to fly around in the scene one time but wasn't able to replicate that. Everything else works as expected.

PR #287, opened 2019-06-25
header logo(MYR) will now redirect to the new scene

Jlu18:


src/components/structural/header/Header.js
```diff
@@ -651,7 +651,9 @@ class Header extends Component {
                             Classrooms
                         </Button>
                     </Sidebar>
-                    <h1 className="mr-2 d-none d-sm-block" >MYR</h1>
+                    <h1 className="mr-2 d-none d-sm-block"
```
jasondkiesling: Looks great. Super minor change. When a block with it's opening and closing tag is more than one line, we like to put them on individual lines to make it easier to read. (The formatting is slightly off, but I put a code snippet below that shows how it should look like.)

```<h1 className="mr-2 d-none d-sm-block"
                        style={{ cursor: "pointer" }}
                        onClick={() => { window.location.href = window.origin; }} >
MYR
</h1>
```

Issue #292, opened 2019-06-25
Add ability to pass array to getRandomColor()


jasondkiesling:
If an array is passed, choose a color at random.
ex:
```
let colors = ["blue","green","red","hotpink","FF00FF"];  
colorShift(box(), getRandomColor(colors));  
```
However, if no array is passed it will keep the current method of generating a random color.

PR #295, opened 2019-06-25
Adds ability to pass array to getRandomColor()

samuel-zuk:
addresses [#292](https://github.com/engaging-computing/MYR/issues/292)


src/myr/Myr.js
```diff
@@ -323,14 +323,20 @@ class Myr {
...
-        let color, i, letters;
-        letters = "0123456789ABCDEF";
-        color = "#";
-        i = 0;
-        while (i < 6) {
-            color += letters[Math.floor(Math.random() * 16)];
-            i++;
+    getRandomColor = (colors) => {
+        let color;
+        if (colors.isArray) {
```
jasondkiesling: Let's also check to make sure that there is at least 1 element in the array.

PR #296, opened 2019-06-25
Uncaught editor error in reference page

jSwanson99:
It seems to be highlighting changes (in view) that are already in dev. Am I doing something wrong/is there a reason for this?

### DIFF
--- a/src/components/structural/header/Header.js
+++ b/src/components/structural/header/Header.js
@@ -49,7 +49,8 @@ class Header extends Component {
             navAwayModal: false,
             needsNewId: false, // this explicitly tells us to make a new id
             spinnerOpen: false,
-            referenceOpen: false
+            referenceOpen: false,
+            editorChange: false,
         };
     }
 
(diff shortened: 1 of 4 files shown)


jasondkiesling:
You had to merge the changes from dev into your local branch. I fixed that. Also, please make sure to include what the issue is in the pull request to make it more clear in the commit history and not require looking up an issue number.

src/components/structural/header/Header.js
```diff
@@ -133,21 +133,24 @@ class Header extends Component {
...
-        try {
-            let editor = window.ace.edit("ace-editor");
-            editor.getSession().on("change", () => {
-                let text = editor.getSession().getValue();
-                if (this.props.text !== text) {
-                    this.setState({ editorChange: true });
-                } else {
-                    this.setState({ editorChange: false });
-                }
+        if (window.location.href !== window.origin + "/reference") {
```
jasondkiesling: Let's do this with layout types instead. Check the reference page to see how it's done there.

jasondkiesling: Actually, it might not be the reference page. It might be in the header. Search for `layout_types.REFERENCE` (I think that's what I named it).

PR #303, opened 2019-06-27
Adds ability to pass array to getRandomColor()

samuel-zuk:
addresses [#292](https://github.com/engaging-computing/MYR/issues/292)
tested more thoroughly this time, and is confirmed to be working.


src/tests/Myr.test.js
```diff
@@ -40,6 +40,13 @@ describe("Updates to Myr's Model", () => {
         expect(color).not.toBeUndefined();
         expect(colorRegEx.test(color)).toBeTruthy();
     });
+
+    it("pick Random Color out of a list of colors", () => {
```
jasondkiesling: This test ensures that a color was set, but not that it was pulled from the array. Can we modify this test to make sure it's from the array?

src/myr/reference.js
```diff
@@ -111,8 +111,8 @@ let transformations = [
         example: "setColor"
     },
     {
-        name: <span>getRandomColor()</span>,
-        description: <span>Changes the color of the cursor to a random color, and returns the color. Default: 'red'</span>,
+        name: <span>getRandomColor(<span className="array">array</span>)</span>,
+        description: <span>Changes the color of the cursor to either a completely random color or a random color from an array of colors if an array is provided, then returns that color. Default: 'red'</span>,
```
jasondkiesling: Please talk to Elena about the wording of this

src/tests/Myr.test.js
samuel-zuk: `expect(colors.includes(color)).toBeTruthy();`
this line should do that, i think

PR #310, opened 2019-07-01
Modify autocomplete to pull functions from reference

Jlu18:
Few notes:
- I didn't pull from the group because it's mostly the same function.
- Function in reference without an example will not be shown in the autocomplete. (eg. increasePosition) 
- Still working on autocomplete to show the information about what each function does.


jasondkiesling:
What was the design decision for `Function in reference without an example will not be shown in the autocomplete. (eg. increasePosition)`?

Jlu18:
It's easier to get the example which is string compare to name where's HTML code

jasondkiesling:
Okay. Elena is working on updating some other things in the reference, I'll have her update the way name is being stored as well to make it easier. We do want this to be all of the functions from the reference.

jasondkiesling:
The new reference format should be in by the end of today for you to update this with to allow you to pull from the name versus the example name. Descriptions aren't quite ready. If you want to open a new pull request that just updates the way we pull things into the autocomplete, feel free to do that.

As this is still a work in progress, I'm going to close it for now. When it's ready for review again, just open a new pull request. If you need help in the meantime, feel free to open a draft pull request.

PR #314, opened 2019-07-02
Text function will now only take string as an argument

Jlu18:


jasondkiesling:
If someone doesn't pass in a string we should show "Default". Right now it is confusing if you pass a number as nothing appears.

PR #332, opened 2019-07-11
Modify autocomplete to pull functions from reference

Jlu18:


src/reducers/editor.js
```diff
@@ -31,7 +31,7 @@ m.init();
 // ESLint doesn't like this but it is better than eval
 function noEvalEvaluation(text) {
     // eslint-disable-next-line
-    return Function(`'use strict'; ${text}`)();
+  return Function(`'use strict'; ${text}`)();
```
jasondkiesling: This should have an indent of 4 spaces.

package.json
```diff
@@ -1,64 +1,64 @@
 {
```
jasondkiesling: This should also be indented 4 spaces.

jasondkiesling:
Approved. Looks like a rebase is needed locally however.

package.json
```diff
@@ -1,64 +1,64 @@
...
-    },
-    "proxy": "http://localhost:1337/",
-    "browserslist": [
-        ">0.2%",
-        "not dead",
-        "not ie <= 11",
-        "not op_mini all"
-    ]
-}
+  "name": "myr",
```
jasondkiesling: Formatting is off again on this.

PR #338, opened 2019-07-15
1.5.5 |  ES Lint Updates, catch firebase errors, firebase scene filtering, and infinite loop detector POC

jasondkiesling:
- Scenes without a preview image now appear
- Firebase scenes with old names and IDs in settings are removed.
- ESLint now doesn't allow != and code is updated to reflect that.
- Infinite loop detector (commented out)


jasondkiesling:
I need to rebase this before it can be merged.

jasondkiesling:
This is now ready for review

PR #376, opened 2019-10-15
API updates

kdvalin:
Frontend courses now work with [API update for Backend](https://github.com/engaging-computing/MYR-backend/pull/142).


jasondkiesling:
One thing to make sure that works is that we handle errors properly. For instance, what if a course has no lessons?

PR #380, opened 2019-12-04
MYR Random Functions

Joshua119:
Functions to create "random" seeded numbers in MYR.

- this.oldRandomCounter and this.oldSeedTime aren't used in the code itself but were added to help the cursor popup generate the same random numbers.


Joshua119:
The reason the functions use both a seed and a counter is to make it possible for the cursor popup to repeat the random numbers. We cant make random numbers normally for the cursor popup because the popup updates not after every line is executed, but every one in a while, presumably with a variable number of lines being run in between. As such, the random numbers not only need to be seeded, but tracked with a counter so the cursor popup can go back and get the correct random numbers no matter how many times random/randomInt are run between cursor popup updates.

src/myr/Myr.js
```diff
@@ -11,6 +11,10 @@ class Myr {
         this.assets = [];
         this.res = { els: this.els, assets: this.assets };
         this.sceneEl = document.querySelector("a-scene");
+        this.seed = 0; 
+        this.randCounter = 0;
+        this.oldRandomCounter = 0;
+        this.oldSeed = 0;
```
jasondkiesling: Similarly to this.cursor, it would be good to have these as a JSON object.

src/myr/Myr.js
```diff
@@ -1047,6 +1051,153 @@ class Myr {
...
+            {
+                seedNum *= seedNum;
+            }
+
+            nextRand = this.nextRandom(min, max);
+            this.decrementRandCounter();
+
+            let small = seedNum*0.000000000000000001;
+            let counterPow = this.randCounter * this.randCounter;
+            randNum = (seedNum*Math.PI*counterPow*small*1.543637*9.479137*0.5*nextRand);
```
jasondkiesling: What's the meaning behind these numbers?

src/myr/Myr.js
```diff
@@ -1047,6 +1051,153 @@ class Myr {
...
+            {
+                seedNum *= seedNum;
+            }
+
+            nextRand = this.nextRandom(min, max);
+            this.decrementRandCounter();
+
+            let small = seedNum*0.000000000000000001;
+            let counterPow = this.randCounter * this.randCounter;
+            randNum = (seedNum*Math.PI*counterPow*small*1.543637*7.479137*0.5*nextRand);
```
jasondkiesling: We seem to be duplicating some logic here. It would be good to combine these with a helper function.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,120 @@
...
+        zPos = Math.floor(zPos);
+        myr.setPosition(1, 1, zPos);
+        if(zPos >= 15 && zPos < 20) {
+            within = 1;
+        }
+        expect(within).toEqual(1);
+    });
+
+    it("incorrect randomInt value should return underfined", () => {
+        let zPos = myr.random(20, 15);
```
jasondkiesling: Is this actually the behavior we want? It would probably be best for new users to swap them for them.

src/myr/Myr.js
Joshua119: changed to being a JSON object

src/myr/Myr.js
Joshua119: There is no specific meaning, they are only meant to make the resulting 'random' number less predictable.

src/myr/Myr.js
Joshua119: turned into helper function calculateRandom.

src/tests/Myr_random.test.js
Joshua119: Changed to swapping the values instead of throwing an error.

jasondkiesling:
Please check your tests. They should not be failing.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
         return outerElId;
     }
 
+    setSeed = (seed = 0) => {
+        if (seed === 0){
```
jasondkiesling: If passing nothing to setSeed should generate a random seed, then you should not pass a default argument and should instead check if seed is undefined.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+    }
+
+    decrementRandCounter = () => {
+        this.rand.randCounter -= 1;
+    }
+
+    calculateRandom = (seedNum, nextRand) => {
+        let small = seedNum*0.000000000000000001;
+        let counterPow = this.rand.randCounter * this.rand.randCounter;
+        return (seedNum*Math.PI*counterPow*small*1.543637*9.479137*0.5*nextRand);
```
jasondkiesling: This code is hard to read and understand. You're throwing a lot of numbers that aren't clear what they are for.

Also, if you're not changing a value it should be const.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+    }
+
+    randomInt = (min = -40, max = 40) => {
+        if(max < min)
+        {
+            let temp = max;
+            min = max;
+            max = temp;
+        }
+        if(max === min)
```
jasondkiesling: This should be an else if.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+        let range = max - min;
+        let randNum, seedNum, nextRand;
+
+        if(this.rand.seed === 0)
+        {
+            this.rand.seed = new Date().getTime();
+            this.rand.randCounter = 0;
+        }
+
+        if(this.rand.seed !== 0)
```
jasondkiesling: This if is not needed. Your check above already made sure that the seed is not 0.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+        }
+        if(max === min)
+        {
+            return max;
+        }
+
+        let range = max - min;
+        let randNum, seedNum, nextRand;
+
+        if(this.rand.seed === 0)
```
jasondkiesling: Is it ever possible to set the seed to 0? If not, then you don't need this check.

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+            }
+
+            nextRand = this.nextRandom(min, max);
+            this.decrementRandCounter();
+
+            randNum = this.calculateRandom(seedNum, nextRand);
+            while(randNum > 100000000000000)
+            {
+                randNum = randNum / 100;
+            }
```
jasondkiesling: Why don't you use modulo for this?

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+        if(this.rand.seed !== 0)
+        {
+            this.rand.randCounter += 1;
+            seedNum = this.rand.seed;
+            while(seedNum < 1000000000)
+            {
+                seedNum *= seedNum;
+            }
+
+            nextRand = this.nextRandom(min, max);
```
jasondkiesling: What is nextRand and this.nextRandom?

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+            this.decrementRandCounter();
+
+            randNum = this.calculateRandom(seedNum, nextRand);
```
jasondkiesling: What is difference between nextRandom and calculateRandom? Why are they separate functions?

src/myr/Myr.js
```diff
@@ -1047,6 +1053,164 @@ class Myr {
...
+                randNum = randNum / 100;
+            }
+            randNum = randNum % range;
+        }
+        randNum = Math.floor(randNum);
+        randNum = randNum + min;
+        return randNum;
+    }
+
+    random = (min = -40, max = 40) => {
```
jasondkiesling: Similar comments to above.

src/myr/reference.js
```diff
@@ -363,6 +363,26 @@ let transformations = [
         description: <span>The setPhiLength function sets the phi length of the cursor in degrees. This changes the number of degrees shown of certain shapes. Interesting shapes can be achieved with values greater than 360. The default phi length is 360.</span>,
         example: "setPhiLength"
     },
+    {
```
jasondkiesling: Are this actually transformations?

src/myr/reference.js
```diff
@@ -363,6 +363,26 @@ let transformations = [
...
+    },    
+    {
+        name: "setSeed",
+        parameters: [{ type: "number", name: "seed" }],
+        description: <span>The setSeed function sets the seed used by the randomInt and random functions to generate random numbers. The default value is based on the current time.</span>,
+    },
+    {
+        name: "getSeed",
+        parameters: [],
+        description: <span>The getSeed function returns the value of the current seed.</span>,
```
jasondkiesling: We should make sure that our descriptions don't say the obvious. There should be more information about this.

src/myr/reference.js
```diff
@@ -363,6 +363,26 @@ let transformations = [
...
     },
+    {
+        name: "randomInt",
+        parameters: [{ type: "number", name: "min" }, { type: "number", name: "max" }],
+        description: <span>The randomInt function generates a random integer derived from the current seed. The range begins with the minimum number and includes every integer less than the maximum. The default range is -40 to 40.</span>,
+    },
+    {
+        name: "random",
+        parameters: [{ type: "number", name: "min" }, { type: "number", name: "max" }],
+        description: <span>The random function generates a random float derived from the current seed. The range begins with the minimum number and includes every number less than the maximum. The default range is -40 to 40.</span>,
```
jasondkiesling: Keep our audience in mind here. For a middle school student, what is a "float"?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+const myr = new Myr();
+
+describe("Random Function Tests", () => {
+    
+    //randomInt
+    it("randomInt should move box to random x coordinate", () => {
+        let xPos = myr.randomInt();
+        myr.setPosition(xPos, 1, 1);
+        expect(myr.cursor.position).toEqual({ x: xPos, y: 1, z: 1 });
+    });
```
jasondkiesling: What is this test actually testing? It's testing setPosition more than randomInt.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+
+    it("randomInt value should be within default range", () => {
+        let within = 0, xPos = myr.randomInt();
+        xPos = myr.randomInt();
+        myr.setPosition(1, 1, xPos);
+        if(xPos >= -40 && xPos < 40) {
+            within = 1;
+        }
+        expect(within).toEqual(1);
+    });
```
jasondkiesling: Without changing the overall code, you should not use 1/0 for true/false in JS. You should be using true/false. There is also extra complexity. Read about how expect works [here](https://jestjs.io/docs/en/expect). There are better functions for checking booleans that toEqual.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        if(xPos >= 15 && xPos < 20){
+            within = 1;
+        }
+        expect(within).toEqual(1);
+    });
+
+    //random
+    it("random should move box to random z coordinate", () => {
+        let zPos = myr.random();
+        zPos = Math.floor(zPos);
```
jasondkiesling: Why are you using random and then calling Math.floor instead of randomInt? Again, what is this actually testing?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        myr.setPosition(1, 1, zPos);
+        expect(myr.cursor.position).toEqual({ x: 1, y: 1, z: zPos });
+    });
+
+    it("random number should be within default range", () => {
+        let within = 0, zPos = myr.random();
+        zPos = Math.floor(zPos);
```
jasondkiesling: Again, why is this using Math.floor?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+    });
+
+    it("random value should be within preset range", () => {
+        let within = 0, xPos = myr.random(15, 20);
+        xPos = Math.floor(xPos);
+        myr.setPosition(1, 1, xPos);
+        if(xPos >= 15 && xPos < 20){
+            within = 1;
+        }
+        expect(within).toEqual(1);
```
jasondkiesling: Same as above.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+    it("random number should equal 0", () => {
+        let zPos = myr.random(5,5);
+        expect(zPos).toEqual(5);
+    });
+
+    //seeding
+    it("getSeedCounter should return correct seed value", () => {
+        myr.setSeed();
+        let seed = myr.getSeed();
+        expect(myr.rand.seed).toEqual(seed);
```
jasondkiesling: This could be simplified to be one line.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        expect(within).toEqual(1);
+    });
+
+    it("random number should equal 0", () => {
+        let zPos = myr.random(5,5);
+        expect(zPos).toEqual(5);
+    });
+
+    //seeding
+    it("getSeedCounter should return correct seed value", () => {
```
jasondkiesling: Please check that you're naming all of your tests based on what they test.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        let within = 0, zPos = myr.random(20, 15);
+        zPos = Math.floor(zPos);
+        myr.setPosition(1, 1, zPos);
+        if(zPos >= 15 && zPos < 20) {
+            within = 1;
+        }
+        expect(within).toEqual(1);
+    });
+
+    it("random number should equal 0", () => {
```
jasondkiesling: Random number should equal 0?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        let seed = myr.getSeed();
+        expect(myr.rand.seed).toEqual(seed);
+    });
+
+    it("setSeed should set seed to given seed", () => {
+        myr.setSeed(5);
+        expect(myr.rand.seed).toEqual(5);
+    });
+
+    it("getSeed should return correct seed", () => {
```
jasondkiesling: You duplicate this test case (one with an improper name)

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+    });
+    
+    //counter 
+    it("counter should increment to 1", () => {
+        myr.setSeed();
+        myr.randomInt();
+        expect(myr.rand.randCounter).toEqual(1);
+    });
+
+    it("getSeedCounter should return correct counter value", () => {
```
jasondkiesling: Again, naming.

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,125 @@
...
+        myr.randomInt();
+        myr.randomInt();
+        myr.decrementRandCounter();
+        myr.decrementRandCounter();
+        expect(myr.rand.randCounter).toEqual(0);
+    });
+
+    it("should return the same 'random' number", () => {
+        myr.setSeed();
+        let first = myr.randomInt();
```
jasondkiesling: Anything that will never change should be const. This goes for all of your tests.

src/myr/reference.js
Joshua119: Its a bit of a stretch I know, but they're more associated with transformations than geometry, animations, or groups. As random and randomInt would likely be used mostly in conjuntion with transformations like setPosition, setScale, and setColor (plus getRandomColors was already in transformations), it seemed like the most natural section of the four if we dont want to make a 5th category.

src/myr/reference.js
Joshua119: Changing it to "The getSeed function returns the value of the current seed used by the randomInt and random functions to generate random numbers." I'm not sure what else to add that would be helpful without going into a long explanation of what seeding is.

src/myr/Myr.js
Joshua119: The seed is first initialized to 0 in the constructor, which is also why several functions check if its 0 instead of uninitialized.

src/myr/Myr.js
Joshua119: this loop is largely to deal with occasional trailing zeros at high counter values which modulo does not deal with.

src/myr/Myr.js
Joshua119: nextRandom gets a pseudo-random number to use in the calculation of our actual random number.

src/myr/Myr.js
Joshua119: nextRandom gets a pseudo-random number. calculateRandom uses this number along with its own calculations to generate the final random number.

src/myr/Myr.js
jasondkiesling: This goes against JS best practices. It shouldn't be initialized to 0 if 0 isn't a valid value. It should be left as undefined and then you can check if it's undefined or not.

src/myr/Myr.js
jasondkiesling: What happens if you pass in a negative? Or a string? Or a JSON object?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,123 @@
...
+describe("Random Function Tests", () => {
+    
+    //randomInt
+    it("randomInt value should be within default range", () => {
+        const rand = myr.randomInt();
+        let within;
+
+        within = Boolean(rand >= -40 && rand < 40);
+
+        expect(within).toBeTruthy();
```
jasondkiesling, line 14: These could be done with 1 line instead of 3

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,123 @@
+import Myr from "../myr/Myr";
+
+const myr = new Myr();
+
+describe("Random Function Tests", () => {
+    
+    //randomInt
```
jasondkiesling, line 7: Can we check that we actually only get integers?

src/tests/Myr_random.test.js
```diff
@@ -0,0 +1,123 @@
...
+        
+        expect(within).toBeTruthy();
+    });
+
+    it("random number should equal 5", () => {
+        const rand = myr.randomInt(5,5);
+        expect(rand).toEqual(5);
+    });
+
+    //random
```
jasondkiesling, line 40: Can we check that we get non-integers?

Joshua119: Added check

src/tests/Myr_random.test.js
Joshua119: Added check

kdvalin:
Closing this as it is stale and no longer worked on.

PR #383, opened 2020-02-14
Images and Scene API Update

kdvalin:
Frontend changes to accommodate backend PR [#144](https://github.com/engaging-computing/MYR-backend/pull/144).

### DIFF
--- a/src/actions/editorActions.js
+++ b/src/actions/editorActions.js
@@ -47,10 +47,12 @@ export function fetchScene(id, uid = "anon") {
                 if(response.status === 404){
                     window.location.href = window.origin + "/error-404";
                 }
+                console.error("Error retrieving scene. Reason: ", response.statusText);
+                return;
             }
             response.json().then((json) =>{
                 if(json.code){
-                    dispatch(render(json.code, uid || "annon"));
+                    dispatch(render(json.code, uid || "anon"));
                     dispatch(updateSavedText(json.code));
                     let settings = DEF_SETTINGS;
 
--- a/src/actions/editorActions.js
+++ b/src/actions/editorActions.js
@@ -1,7 +1,8 @@
-import { scenes } from "../firebase.js";
 import { loadScene } from "./sceneActions";
 import { DEF_SETTINGS } from "../reducers/scene";
 import * as types from "../constants/ActionTypes";
+
+const sceneRef = "/apiv1/scenes";
 
 /**
  * @function - Sends a signal to the reducer to render the scene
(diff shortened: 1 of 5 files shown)


src/actions/editorActions.js
```diff
@@ -41,41 +42,31 @@ export function recover() {
...
-
-                // if the incoming scene has a settings, merge default with incoming
-                if (data.settings) {
-                    settings = { ...settings, ...data.settings };
+        fetch(`${sceneRef}/id/${id}`).then((response) =>{
+            if(response.status !== 200){
+                if(response.status === 404){
+                    window.location.href = window.origin + "/error-404";
                 }
+            }
```
jasondkiesling: Is there a reason why we're checking for not 200 and then for 404 instead of checking 404 directly? Or should we be checking for more than just 404?

src/actions/editorActions.js
```diff
@@ -41,41 +42,31 @@ export function recover() {
...
+            response.json().then((json) =>{
+                if(json.code){
+                    dispatch(render(json.code, uid || "annon"));
```
jasondkiesling: This should be "anon".

src/actions/editorActions.js
```diff
@@ -41,41 +42,31 @@ export function recover() {
...
-                // Use default for eventual consistency in db
-                let settings = DEF_SETTINGS;
-
-                // if the incoming scene has a settings, merge default with incoming
-                if (data.settings) {
-                    settings = { ...settings, ...data.settings };
+        fetch(`${sceneRef}/id/${id}`).then((response) =>{
+            if(response.status !== 200){
+                if(response.status === 404){
+                    window.location.href = window.origin + "/error-404";
```
jasondkiesling: Did you check what happens with an invalid ID?

src/actions/projectActions.js
```diff
@@ -1,33 +1,22 @@
...
-                            num_proj_loaded++;
-                            if(num_proj_loaded === snap.size){
-                                dispatch(syncUserProj(userVals));
-                            }
-                        });
-                });
+            fetch(`${sceneRef}/`, {headers: {"x-access-token": id}}).then((response) =>{
+                if(response.status === 200){
+                    response.json().then((json) =>{
+                        for(let i = 0; i < json.length; ++i){
```
jasondkiesling: It's better to use .map or a .forEach than a for loop for JS arrays. It helps avoid issues such as off by one errors.

jasondkiesling: If you didn't know, you can get the index value in .map for a .forEach. The callback has an optional arg of index.

src/actions/projectActions.js
```diff
@@ -40,29 +29,20 @@ export function syncUserProj(payload) {
...
-                        let dat = doc.data();
-                        exampleVals.push({
-                            name: dat.name,
-                            id: doc.id,
-                            data: dat,
-                            url: img ? img : "/img/no_preview.jpg"
+        fetch(`${sceneRef}/`, {headers: {"x-access-token": "1"}}).then((response) =>{
+            if(response.status === 200){
+                response.json().then((json) =>{
+                    for(let i = 0; i < json.length; ++i){
```
jasondkiesling: Again here with the for loop.

src/actions/projectActions.js
```diff
@@ -40,29 +29,20 @@ export function syncUserProj(payload) {
...
+                        let id = json[i].firebaseID ? json[i].firebaseID : json._id;
```
jasondkiesling: The ideal thing to do here would be to have the migration script change all of the image names to the new mongo id. This way we don't have any conflicts later.

src/actions/editorActions.js
kdvalin: The backend returns the "Not found" image if the ID doesn't exist/doesn't have an image

src/actions/editorActions.js
kdvalin: Forgot to check for other errors, fixing in upcoming commit

src/actions/projectActions.js
kdvalin: Will update, seems I also forgot to handle the example scene images, since they're still being grabbed from firebase.

kdvalin:
Closing since full firebase migration branch will include these changes

PR #385, opened 2020-02-21
Vertical Flight Keys!

samuel-zuk:


src/utils/WASDPlusControls.js
```diff
@@ -0,0 +1,207 @@
...
+    "38": "ArrowUp",
+    "37": "ArrowLeft",
+    "40": "ArrowDown",
+    "39": "ArrowRight",
+    "87": "KeyW",
+    "65": "KeyA",
+    "83": "KeyS",
+    "68": "KeyD",
+    "32": "Space",
+    "16": "ShiftLeft"
```
jasondkiesling: Should this have ShiftRight?

src/utils/WASDPlusControls.js
```diff
@@ -0,0 +1,207 @@
+import AFRAME from "aframe";
+import * as THREE from "three";
+let bind = AFRAME.utils.bind;
+let shouldCaptureKeyEvent = AFRAME.utils.shouldCaptureKeyEvent;
+
+let KEYCODE_TO_CODE = {
```
jasondkiesling: This should be const.

src/utils/WASDPlusControls.js
```diff
@@ -0,0 +1,207 @@
...
+    "39": "ArrowRight",
+    "87": "KeyW",
+    "65": "KeyA",
+    "83": "KeyS",
+    "68": "KeyD",
+    "32": "Space",
+    "16": "ShiftLeft"
+};
+
+let KEYS = [
```
jasondkiesling: This should be const. It should also probably be generated from the above object.

src/utils/WASDPlusControls.js
```diff
@@ -0,0 +1,207 @@
...
+    "KeyW", "KeyA", "KeyS", "KeyD",
+    "ArrowUp", "ArrowLeft", "ArrowRight", "ArrowDown",
+    "Space", "ShiftLeft", "ShiftRight"
+];
+
+let MAX_DELTA = 0.25;
+let CLAMP_VELOCITY = 0.01;
+let MOD_Y = 1.5;
```
jasondkiesling: Do these ever change? If so, they should be const.

src/utils/WASDPlusControls.js
samuel-zuk: So this object, which was a direct copy from A-Frame's wasd controls component, appears to only be used for backwards compatibility with the KeyboardEvent.keyCode value which has since been deprecated by Mozilla. (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode). I'm thinking it's best to just drop this object altogether, actually.

src/utils/WASDPlusControls.js
```diff
@@ -0,0 +1,193 @@
...
+    "KeyW", "KeyA", "KeyS", "KeyD",
+    "ArrowUp", "ArrowLeft", "ArrowRight", "ArrowDown",
+    "Space", "ShiftLeft", "ShiftRight"
+];
+
+const MAX_DELTA = 0.25;
+const CLAMP_VELOCITY = 0.01;
+const MOD_Y = 1.5;
+
+let isEmptyObject = (keys) => {
```
jasondkiesling: Sorry I missed this earlier. This function should also be const.

src/components/structural/header/MyrTour.js
```diff
@@ -55,25 +55,49 @@ class MyrTour extends Component {
...
-        content: "The Stop button will stop the scene. \n Use this to save battery."
+        content: lineBreakHelper(stop)
     },
     {
         selector: "#scene",
-        content: "The View is where you can see your work. \n Click the goggle to view in VR."
+        content: lineBreakHelper(view)
     },
     {
         selector: "#new-btn",
```
jasondkiesling, line 103: Can you think of a way to improve the readability of this? I'm not the biggest fan of having two different standards for where the content is stored. Perhaps it would make sense to create a new JSON object with all of the data, and then generate steps by stepping through that object.

jasondkiesling: The more I think about it, the more I know. Let's do this how we do with our autocomplete. Make a JSON object array of the steps. Then map each element of the array into the steps array that gets passed into the autocomplete that calls the formatter on every content item. This way there is one variable with all of the data.

jasondkiesling:
There's just that one thing that needs to be fixed. Once that's done, this is ready to be deployed to beta for testing!

PR #390, opened 2020-04-07
Textures

Joshua119:
setTexture function added and works with both keys and URL's, initial texture.js textures added.


src/components/structural/Textures.js
```diff
@@ -0,0 +1,30 @@
...
+    {pavement : "https://static.turbosquid.com/Preview/2014/08/01__17_56_46/pavement%201b.jpgba228c77-b71f-4636-a8d9-6da2dde724dbLarger.jpg"},
+    {sand : "https://upload.wikimedia.org/wikipedia/commons/thumb/5/5a/Sand.jpg/1024px-Sand.jpg"},
+    {snow : "https://upload.wikimedia.org/wikipedia/commons/0/0b/Shadows_on_snow.jpg"},
+    {steel : "https://img.freepik.com/free-photo/silver-metal-texture-background-design_46250-2268.jpg?size=626&ext=jpg"},
+    {stars : "https://upload.wikimedia.org/wikipedia/commons/b/b5/Graphic_patterns_by_Peak_Hora_89.jpg"},
+    {water : "https://upload.wikimedia.org/wikipedia/commons/thumb/3/3b/Mink_Mingle_2017-04-23_%28Unsplash_AlEyuIKuSmc%29.jpg/1200px-Mink_Mingle_2017-04-23_%28Unsplash_AlEyuIKuSmc%29.jpg"},
+    {wicker : "https://upload.wikimedia.org/wikipedia/commons/d/da/Basketwork_wicker_basket_close_up.jpg"},
+    {wood : "https://img.freepik.com/free-photo/wooden-textured-background_53876-14865.jpg?size=626&ext=jpg"},
+    {wool : "https://upload.wikimedia.org/wikipedia/commons/thumb/3/35/Beige_wool_texture.jpg/1200px-Beige_wool_texture.jpg"},
+];
```
jasondkiesling, line 28: We shouldn't be relying on outside sources to host these images. We should host them ourselves instead. We can then also optimize the file sizes as we choose.

src/components/structural/View.js
```diff
@@ -117,6 +117,7 @@ class View extends Component {
     assetsHelper = (asset) => {
         return (
             <a-asset-item key={asset.id} id={asset.id} src={asset.src}></a-asset-item>
+
```
jasondkiesling, line 120: What is this line?

src/myr/Myr.js
```diff
@@ -11,8 +12,11 @@ class Myr {
         this.assets = [];
         this.res = { els: this.els, assets: this.assets };
         this.sceneEl = document.querySelector("a-scene");
+        this.seed = 0; 
+        this.oldSeed = 0;
```
jasondkiesling, line 16: The math stuff shouldn't be in this pull request.

src/myr/Myr.js
```diff
@@ -421,7 +427,25 @@ class Myr {
...
+    setTexture = (texture = "blank") => {
+        let holdKey;
+
+        TexturePack.forEach(function(item) {
+            Object.keys(item).forEach(function(key) {
+                if(key === texture){
+                    holdKey = item[key];
+                }
+            });
+        });
```
jasondkiesling, line 439: You should change your data type in TexturePack. This logic is way too complex for what you're doing. Your data type is also not following good practices. There are 2 options on how to do so:
```js
TexturePack={
"bark":"url.com/bark.jpg",  
"brick":"url.com/brick.jpg"  
}
```

or

```js
TexturePack=[  
{title:"bark",url:"url.com/bark.jpg"},  
{title:"brick",url:"url.com/brick.jpg"}  
]
```

src/myr/Myr.js
```diff
@@ -421,7 +427,25 @@ class Myr {
...
+        if(holdKey){
+            this.cursor.texture = holdKey;
+        } else {
+            this.cursor.texture = texture;
```
jasondkiesling, line 443: We should have at least some form of basic regex check to make sure the URL is valid.

src/myr/reference.js
```diff
@@ -253,6 +253,11 @@ let transformations = [
         description: <span>The setColor function changes the color of the cursor. For general colors, <a href='https://htmlcolorcodes.com/color-names/' target="_blank" rel="noopener noreferrer">HTML color codes</a> can be used (e.g., "blue", "deeppink"), and for specific colors HEX colors can be used (e.g., "#00ffff"). The default cursor state is "red".</span>,
         example: "setColor"
     },
+    {
+        name: "setTexture",
+        parameters: [{type: "string", name: "texture"}],
+        description: <span>The setColor function changes the texture of the cursor. The unaltered texture is displayed when the color is white, otherwise the texture will be affected by the current color.</span>,
```
jasondkiesling, line 259: How will an end user know our texture options?

Joshua119: Is there anywhere you think it would be best to list them? Textures will now autocomplete which should also help.

src/components/structural/Textures.js
Joshua119: Where would be the best place to host them?

src/components/editor/customCompleter.js
```diff
@@ -192,6 +192,44 @@ export const customCompleter = {
...
+            "pavement",
+            "sand",
+            "snow",
+            "steel",
+            "stars",
+            "water",
+            "wicker",
+            "wood",
+            "wool"
+        ];
```
jasondkiesling, line 222: You should never repeat yourself in code. These should be pulled programmatically from the same place where they're used for the textures.

src/components/structural/Textures.js
```diff
@@ -0,0 +1,30 @@
+let TexturePack = [
```
jasondkiesling, line 1: My original comment still stands. These should be hosted by us, not by a third party.

src/myr/Myr.js
jasondkiesling: There should be some form of error handling here if it's invalid.

src/myr/Myr.js
```diff
@@ -421,7 +428,20 @@ class Myr {
         return this.cursor.color;
     }
 
-    getRandomColor = (colors = null) => {
+    setTexture = (texture = "blank") => {
+        for (let i=0; i < TexturePack.length; i++) {
+            if (TexturePack[i].title === texture) {
+                this.cursor.texture = TexturePack[i].url;
+            }
+        }
```
jasondkiesling, line 436: There are much better ways of doing this. In JS, you should almost never use a for loop with an array. Look up how to use indexOf. It's also probably best to return as soon as you set the texture as you're doing extra work you don't have to.

src/myr/Myr.js
```diff
@@ -514,10 +534,11 @@ class Myr {
         let base = {
             geometry: "primitive: box;",
             id: "box" + this.genNewId(),
-            material: `color: ${this.cursor.color};`,
+            material: `color: ${this.cursor.color}; side: double; src: ${this.cursor.texture};`,
```
jasondkiesling, line 537: These should be consistent across all geometries as to what order the material properties are being set.

src/myr/Myr.js
```diff
@@ -530,7 +551,7 @@ class Myr {
             position: this.cursor.position,
             scale: this.cursor.scale,
             rotation: this.cursor.rotation,
-            material: `color: ${this.cursor.color}; side: double;`,
+            material: `color: ${this.cursor.color}; src: ${this.cursor.texture}; side: double;`,
```
jasondkiesling, line 554: See above comment.

src/myr/Myr.js
```diff
@@ -723,7 +744,7 @@ class Myr {
             position: this.cursor.position,
             scale: this.cursor.scale,
             rotation: this.cursor.rotation,
-            material: `color: ${this.cursor.color};`,
+            material: `color: ${this.cursor.color}; side: double; src: ${this.cursor.texture};`,
```
jasondkiesling, line 747: See above comment.

src/myr/reference.js
```diff
@@ -253,6 +253,11 @@ let transformations = [
         description: <span>The setColor function changes the color of the cursor. For general colors, <a href='https://htmlcolorcodes.com/color-names/' target="_blank" rel="noopener noreferrer">HTML color codes</a> can be used (e.g., "blue", "deeppink"), and for specific colors HEX colors can be used (e.g., "#00ffff"). The default cursor state is "red".</span>,
         example: "setColor"
     },
+    {
+        name: "setTexture",
+        parameters: [{type: "string", name: "texture"}],
+        description: <span>The setTexture function changes the texture of the cursor. The unaltered texture is displayed when the color is white, otherwise the texture will be affected by the current color. Textures can either be applied using the name of one of these /*insert link here*/ colors (e.g. "brick") or by inserting a valid url (e.g. "https://upload.wikimedia.org/wikipedia/commons/5/5e/Water_drop_001.jpg")</span>,
+    },
```
jasondkiesling, line 260: What about getTexture?

src/myr/reference.js
```diff
@@ -253,6 +253,11 @@ let transformations = [
         description: <span>The setColor function changes the color of the cursor. For general colors, <a href='https://htmlcolorcodes.com/color-names/' target="_blank" rel="noopener noreferrer">HTML color codes</a> can be used (e.g., "blue", "deeppink"), and for specific colors HEX colors can be used (e.g., "#00ffff"). The default cursor state is "red".</span>,
         example: "setColor"
     },
+    {
+        name: "setTexture",
+        parameters: [{type: "string", name: "texture"}],
+        description: <span>The setTexture function changes the texture of the cursor. The unaltered texture is displayed when the color is white, otherwise the texture will be affected by the current color. Textures can either be applied using the name of one of these /*insert link here*/ colors (e.g. "brick") or by inserting a valid url (e.g. "https://upload.wikimedia.org/wikipedia/commons/5/5e/Water_drop_001.jpg")</span>,
```
jasondkiesling, line 259: Where are you planning on having that link be?

jasondkiesling (dismissed):
In addition to these comments, there are also two major issues.

1. The existing tests are failing. You should **always run tests** before asking for code review. I will not be doing more code review until all tests are passing. You should run tests locally with `yarn test`.  
1. You did not write any tests for your new use cases. All new or changed functions should have tests.

src/myr/reference.js
Joshua119: No other getters are currently documented in the reference, so I did not document getTexture. would you like me to create references to the other getters along with getTexture while im here?

jasondkiesling:
Converted to a draft pull request until ready for further review

jasondkiesling:
You need to merge the latest changes from master into this.

src/components/structural/Textures.js
```diff
@@ -0,0 +1,28 @@
+let TexturePack = [
+    {title:"bricks",url:"img/textures/bricks.jpg"},
```
jasondkiesling, line 2: Small edit, but can these have a slash at the start. Like "/img/textures/bricks.jpg"?

src/myr/Myr.js
```diff
@@ -421,6 +430,26 @@ class Myr {
...
+        let textures = TexturePack();
+        let textureTitle = [...textures.TexturePack.map(obj => obj.title)];
+        let textureURL = [...textures.TexturePack.map(obj => obj.url)];
+
+        this.cursor.texture = textureURL[textureTitle.indexOf(texture)];
+
+        let urlregex = /^(https?|ftp):\/\/([a-zA-Z0-9.-]+(:[a-zA-Z0-9.&%$-]+)*@)*((25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])){3}|([a-zA-Z0-9-]+\.)*[a-zA-Z0-9-]+\.(com|edu|gov|int|mil|net|org|biz|arpa|info|name|pro|aero|coop|museum|[a-zA-Z]{2}))(:[0-9]+)*(\/($|[a-zA-Z0-9.,?'\\+&%$#=~_-]+))*$/;
+        if(urlregex.test(texture) || (texture === "")) {
+            this.cursor.texture = texture;
+        }
```
kdvalin, line 443: What would happen if the URL fails the regex?  Does it notify the user in some way?

Joshua119: No texture is displayed. The user currently isn't notified, I'm adding a console.error statement. Two probably best solutions would either be an error popup or and error texture however just because something passes the regex doesn't mean a texture will be read/displayed as it can get blocked by the CORS policy when the scene is run and tries to load the URL. So there would be some confusion as to why sometimes an error is shown and sometimes one is not despite neither working.

kdvalin (approved):

kdvalin (approved):

PR #401, opened 2020-06-01
Firebase backend

kdvalin:
Changes for the frontend to work with backend pull request [#149](https://github.com/engaging-computing/MYR-backend/pull/149)


src/actions/collectionActions.js
```diff
@@ -1,23 +1,20 @@
-import { collections, scenes } from "../firebase.js";
```
jasondkiesling, line 1: We shouldn't need the file "firebase.js" anymore, right?

src/actions/editorActions.js
```diff
@@ -41,41 +42,41 @@ export function recover() {
...
-                dispatch(updateSavedText(data.code));
-                // Use default for eventual consistency in db
-                let settings = DEF_SETTINGS;
 
-                // if the incoming scene has a settings, merge default with incoming
-                if (data.settings) {
-                    settings = { ...settings, ...data.settings };
+            if(response.status !== 200){
+                if(response.status === 404){
+                    window.location.href = window.origin + "/error-404";
```
jasondkiesling, line 54: This is probably pulling from old code, but it's better to use window.location.assign(). It's not a big difference, but it's more clear that you're changing the URL and going somewhere.

src/actions/projectActions.js
```diff
@@ -1,33 +1,21 @@
...
-                            });
-                            num_proj_loaded++;
-                            if(num_proj_loaded === snap.size){
-                                dispatch(syncUserProj(userVals));
-                            }
+            fetch(`${sceneRef}/`, {headers: {"x-access-token": id}}).then((response) =>{
+                if(response.status === 200){
+                    response.json().then((json) =>{
+                        json.forEach(element => {
+                            element.url = `${previewRef}/${element._id}`;
```
jasondkiesling, line 14: Hopefully I catch this on the backend too, but can we keep this consistent with our other API calls? So `/apiv1/preview/id/:id`.

src/actions/projectActions.js
```diff
@@ -40,29 +28,15 @@ export function syncUserProj(payload) {
...
-                            data: dat,
-                            url: img ? img : "/img/no_preview.jpg"
-                        });
-                        num_proj_loaded++;
-                        if(num_proj_loaded === snap.size){
-                            dispatch(syncExampleProj(exampleVals));
-                        }
-                    });   
-            });
+        fetch(`${sceneRef}/`, {headers: {"x-access-token": "1"}}).then((response) =>{
```
jasondkiesling, line 31: Again this is a backend comment on frontend code, but can we make an endpoint for examples that doesn't take a token? Such as `/apiv1/scenes/examples`

src/actions/projectActions.js
```diff
@@ -71,29 +45,84 @@ export function syncExampleProj(payload) {
...
+                        fetch(`${sceneRef}/id/${id}`, {method: "delete", headers: {"x-access-token": uid}}).then((response) => {
+                            if(response.status !== 204){
+                                console.error("Error removing document: ", response.status);
+                            }
+                            // If deleting current project, redirect to home
+                            else if (window.location.href === window.origin + "/scene/" + id || window.location.href === window.origin + "/scene/" + id + "/") {
+                                window.location.href = window.origin;
+                            }
+                        }).catch((error) => {
+                            console.error("Error removing document: ", error);
```
jasondkiesling, line 70: If we can't delete the scene, we should directly alert the user.

src/components/structural/header/Header.js
```diff
@@ -52,8 +54,18 @@ class Header extends Component {
             referenceOpen: false,
             coursesOpen: false,
             tourOpen: false,
-            welcomeOpen: false
+            welcomeOpen: false,
+            updateCollect: false,
+            fetchCollect: false,
```
jasondkiesling, line 59: Can we write out the full word "collection"? When I think of collect I think of the verb.

src/components/structural/header/Header.js
```diff
@@ -198,38 +179,55 @@ class Header extends Component {
...
-                        style={{
-                            color: "white",
-                            margin: 4,
-                            padding: 2,
-                            border: "1px solid #fff"
-                        }}>
-                        Log In
-                    </Button>
+                    <GoogleLogin
+                        clientId="12320757914-puvjd9tbs79sc19sl433gqp9sci1tfqt.apps.googleusercontent.com"
```
jasondkiesling, line 209: We should keep this out of source control and use some form of environment variable for it.

PR #402, opened 2020-06-02
Remove case sensitivity from setColor

drpatel44:
I was able to edit the setColor function by adding .toLowerCase so it allows the user to enter both uppercase and lowercase letters and automatically sets the capital letters to lowercase internally.

### DIFF
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -417,7 +417,7 @@ class Myr {
     };
 
     setColor = (color) => {
-        this.cursor.color = color;
+        this.cursor.color = color.toLowerCase();
         return this.cursor.color;
     }
 


src/tests/Myr.test.js
```diff
@@ -853,4 +853,19 @@ describe("Other Myr functionality", () => {
         expect(response).toEqual({ "test1": 1, "test2": 3 });
     });
 
+    it("Should accept any caps or lowercase letters", () => {
```
kdvalin, line 856: You should do some more testing with other colors besides just red.
Red is the default color for MYR, so try some other colors before testing red.

src/tests/Myr.test.js
```diff
@@ -853,4 +853,19 @@ describe("Other Myr functionality", () => {
...
+        myr.setColor("Red");
+        expect(myr.cursor.color).toEqual("red");
+        myr.setColor("rED");
+        expect(myr.cursor.color).toEqual("red");
+        myr.setColor("ReD");
+        expect(myr.cursor.color).toEqual("red");
+    });
+
+    it("Should accept hex colors entered for setColor function", () => {
```
kdvalin, line 865: Add some test cases where letters in the hex color string are capital, like #F008Ab.

PR #405, opened 2020-06-16
v1.9.1 | Bug fixes and case insensitivity for setColor()

jasondkiesling:


jasondkiesling:
@drpatel44 I didn't catch this in your first pull request, but `setColor()` should be able to be handled (without any arguments). The behavior should be similar to how it is now on learnmyr.org.

cc: @kdvalin

PR #414, opened 2020-08-24
Import Export

kdvalin:
This PR adds the ability to export a scene onto a user's computer, and allows that export to be imported on a different account or enviornment.


src/components/structural/header/Header.js
```diff
@@ -550,6 +552,47 @@ class Header extends Component {
...
+        }
+
+        fetch(`/apiv1/scenes/export${param}`, {headers: {"x-access-token": this.props.user.uid}}).then(async (resp) => {
+            switch(resp.status) {
+                case 204:
+                    alert("There are no scenes saved for you to export!");
+                    break;
+                case 200:
+                    let data = await resp.json();
+                    saveAs(new Blob([JSON.stringify(data)]), "MYR-export.json", {
```
jasondkiesling, line 568: We should run the import/export itself in the back end and just use the front end to pass files back and forth.

kdvalin: OK, I will have to restructure the import system to do this (not that bad).

Export has to be that way, the actual creation of JSON is created on the backend, but when invoked from the frontend it does not give a "Save File" prompt (I think this is a problem with `fetch` skipping past the `Content-Disposition` header)

src/components/structural/header/Header.js
```diff
@@ -553,6 +555,43 @@ class Header extends Component {
         this.setState({ snackOpen: false });
     }
 
+    onExport = (scene = undefined) => {
```
samuel-zuk, line 558: I noticed that some, but not all functions in the header have brief inline JSDoc-style documentation, should this be added for these new functions?

samuel-zuk (changes requested):
All the JS code looks good to me on its own but documentation should probably be added to the new helper functions you've added.

Jlu18:
You should add export option in the "SHARING" tab in scene setting to make it consistent.

src/components/structural/header/Header.js
```diff
@@ -553,6 +555,43 @@ class Header extends Component {
...
+        }
+
+        fetch(`/apiv1/scenes/export${param}`, {headers: {"x-access-token": this.props.user.uid}}).then(async (resp) => {
+            switch(resp.status) {
+                case 204:
+                    alert("There are no scenes saved for you to export!");
+                    break;
+                case 200:
+                    let data = await resp.json();
+                    saveAs(new Blob([JSON.stringify(data)]), "MYR-export.json", {
```
Jlu18, line 571: I think having exported file to have a unique name like scene (eg. "MYR-Box.json") might be better for user to recognize what scene is it, but I can see there might be a problem if user have some obscure scene name. (Also realized that you put every scenes in single json file when choose export all)

Jlu18: Also, I'm not a huge fan of putting every scene in single json file when "export all" since user might want to select what scene they want to import. Obviously, it's easy to separate scenes to individual json file but it might not be intuitive for new user. But I think it's fine for now.

Jlu18 (changes requested):
You should add export option in the "SHARING" tab in scene setting to make it consistent.

src/components/structural/header/Header.js
kdvalin: The way that I will have this working is in the "Open" dialog, it exports all the scenes (we can impelment a checkbox thing later if we want), but in the `Sharing` tab it will only export the current scene.

As for the name of the file...  I don't know of too many ways to make it more unique.. I could use the window title, but then it's not really descriptive of the scenes when the user exports all scenes.  Maybe a date appended onto the end?

Jlu18 (dismissed):

kdvalin:
Closing this since the conflicts are really annoying, and it has been rewritten based off a more recent dev branch

PR #416, opened 2020-09-08
Object transparency

kdvalin:
This adds the ability to create translucent objects in MYR.

### DIFF
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -13,7 +13,7 @@ class Myr {
         this.sceneEl = document.querySelector("a-scene");
         this.cursor = {
             color: "red",
-            opacity: 1,
+            transparency: 0,
             position: {
                 x: 0,
                 y: 0,
@@ -81,7 +81,7 @@ class Myr {
         this.id = 0;
         this.cursor = {
             color: "red",
-            opacity: 1,
+            transparency: 0,
             position: {
                 x: 0,
                 y: 0,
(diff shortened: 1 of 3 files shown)


src/myr/Myr.js
```diff
@@ -154,6 +157,14 @@ class Myr {
         return this.counter++;
     };
 
+    setTransparency = (opacity = 0) => {
+        if(typeof opacity === "number" && opacity <= 100 && opacity >= 0) {
+            this.cursor.opacity = (100 - opacity) / 100;
+        }
```
jasondkiesling, line 163: We should always handle errors in some manner. Even if that's a console.error.

src/tests/Myr.test.js
```diff
@@ -880,4 +881,18 @@ describe("Other Myr functionality", () => {
         expect(myr.cursor.color).toEqual("red");
     });
 
+
+    it("setTransparency should set the appropriate cursor attribute correctly", () => {
```
jasondkiesling, line 885: We should always test for the negatives as well (aka make sure stuff fails properly).

jasondkiesling (changes requested):
I can't comment on lines you didn't change, but a case needs to be added to setCursorAttribute as well.

src/myr/reference.js
```diff
@@ -283,6 +283,12 @@ let transformations = [
         description: <span>The setZPos function changes the z component of the position in the cursor. The default z position is 0.</span>,
         example: "setZPos"
     },
+    {
+        name: "setTransparency",
+        parameters: [{type: "number", name: "transparentcy"}],
```
jasondkiesling, line 288: "transparentcy"

Also, do you have a reference example for this?

jasondkiesling: Other than that, this is ready to ship.

jasondkiesling (approved):

Issue #441, opened 2021-02-12
please change default for new project: floor off, grid on


fgmart:
Please change the settings in MYR so that new projects come up with the floor off and the grid on.

This is important so that new users see the grid and the axes (per below).

Thank you!

![Screen Shot 2021-02-12 at 12 30 41 PM](https://user-images.githubusercontent.com/53866836/107801577-3fb9da80-6d2e-11eb-82cc-2fe6da61e60f.png)

kdvalin:
Impelmented as of 2.3.0

PR #443, opened 2021-02-12
Modify lesson so each lesson will load the settings from backend

Jlu18:

PR #451, opened 2021-02-24
Default Scene Settings Update

kdvalin:
This updates the defaults for scenes to show the coordinate grid helper, and hides the floor.

This will close issue #441


kdvalin:
Currently this PR overrides settings for course lessons as well, and there is no real difference between lessons and projects in the frontend.  This will probably be fixed once #443 and it's related PR from the backend are merged into `master`.

kdvalin:
Note, this will affect courses that do not have settings in their lessons.

kdvalin:
> Note, this will affect courses that do not have settings in their lessons.

This should no longer be a problem since as of v2.2.0, all courses have been given settings.

samuel-zuk (approved):

Jlu18 (approved):

PR #459, opened 2021-03-05
Remove cursor update from getRandomColor and add setRandomColor function

Jlu18:
We also have to replace the code in the courses and user scene.

### DIFF
--- a/src/tests/Myr.test.js
+++ b/src/tests/Myr.test.js
@@ -138,11 +138,25 @@ describe("Updates to Myr's Model", () => {
         expect(colorRegEx.test(color)).toBeTruthy();
     });
 
+    it("to set Random Color", ()=> {
+        myr.cursor.color = null;
+        myr.setRandomColor();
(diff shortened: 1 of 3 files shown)


src/tests/Myr.test.js
```diff
@@ -138,12 +138,26 @@ describe("Updates to Myr's Model", () => {
         expect(colorRegEx.test(color)).toBeTruthy();
     });
 
+    it("to set Random Color", ()=> {
+        myr.cursor.color = null;
+        myr.setRandomColor();
+        expect(myr.cursor.color).not.toBeNull(); 
+    });
+
     it("pick Random Color out of a list of colors", () => {
```
kdvalin, line 147: The tests for `getRandomColor` should check to make sure that the color attribute of the cursor object does not update

kdvalin (changes requested):

kdvalin (approved):

PR #490, opened 2021-06-03
Added movement speed slider in scene config menu

samuel-zuk:


kdvalin:
Should the move speed persist throughout reloads/different people viewing the scene?

If so, the backend schema will need to be updated so it can return the movement speed in API requests.

samuel-zuk:
@kdvalin Do we store default user settings? I feel like this would be something we'd want to retain per-user.

kdvalin (approved):

kdvalin:
> @kdvalin Do we store default user settings? I feel like this would be something we'd want to retain per-user.

Negative, that is a feature outside the scope of this PR.


Are we saving the speed on the scene or are we defaulting to the normal speed when someone loads a scene?  Depending on the scale of the scene the default speed could end up being annoying.

PR #491, opened 2021-06-07
Fixed VR icon was always on top of the screen

Jlu18:


samuel-zuk (approved):
Literally just seeing this after implementing the same fix on my end, lol.

kdvalin (approved):

PR #492, opened 2021-06-09
Keyboard Shortcut

Jlu18:
We might need to discuss the description of the shortcuts.
This is how it looks like. 

2021-06-09-115757_927x351_scrot


Jlu18:
replace Cmd with ⌘ (U+2318)

kdvalin (approved):

PR #496, opened 2021-06-26
Add ability to change entity materials and material properties

beewyka819:
Adds the ability for the user to alter the material of any entity between basic, matte, specular, physical, and toon. Furthermore, the roughness and metalness values can be independently altered for entities with physical material.


src/components/structural/View.js
```diff
@@ -137,9 +137,10 @@ class View extends Component {
                 return <a-text key={ent.id} {...flattened}></a-text>;
             }
             if (ent.tube) {
-                return <a-tube path={ent.path} radius={ent.radius} material={ent.material} shadow={shadow} shadowcustomsetting></a-tube>;
+                return <a-tube path={ent.path} radius={ent.radius} material={ent.material} materialtype={`type:${ent.materialtype}`} shadow={shadow} shadowcustomsetting></a-tube>;
```
samuel-zuk, line 140: is there a specific reason why "materialtype={ent.materialtype}" doesn't work here as it does for raidus, material, etc?

src/components/structural/View.js
```diff
@@ -137,9 +137,10 @@ class View extends Component {
...
             }
-            return <a-entity class="raycastable" key={ent.id} {...flattened} shadow={shadow} shadowcustomsetting ></a-entity>;
+            return <a-entity class="raycastable" key={ent.id} {...flattened} material={ent.material} materialinfo={`type: ${ent.materialinfo.type}; roughness: ${ent.materialinfo.roughness}; metalness: ${ent.materialinfo.metalness};`} shadow={shadow} shadowcustomsetting ></a-entity>;
+
```
samuel-zuk, line 143: see above ^^

src/myr/Myr.js
```diff
@@ -670,7 +680,64 @@ class Myr {
...
+            throw new Error("Metalness must be between 0.0 and 1.0!");
+        }
+        this.cursor.metalness = metalness;
+        return this.cursor.metalness;
+    }
+
+    /**
+     * @returns the list of material types
+     */
+    MaterialType = () => {
```
samuel-zuk, line 738: is there a reason why this needs to be a function?

src/myr/Myr.js
```diff
@@ -94,6 +98,9 @@ class Myr {
         this.cursor = {
             color: "red",
             texture: "",
+            materialtype: MaterialType.SPECULAR,
```
samuel-zuk, line 101: (this applies to wherever cursor.materialtype is mentioned) this should probably be camel case (i.e. "materialType") to be consistent with other variables like "textureColoring"

src/utils/AFramePhysics.js
```diff
@@ -80,6 +81,50 @@ AFRAME.registerComponent("shadowcustomsetting", {
     },
 });
 
+AFRAME.registerComponent("materialinfo", {
```
samuel-zuk, line 84: this should probably be placed in a separate file as it's not part of the a-frame physics system. (see https://github.com/engaging-computing/MYR/blob/master/src/utils/AframeRegIndicator.js for example)

src/myr/Myr.js
```diff
@@ -670,7 +680,64 @@ class Myr {
...
+            validMaterial = MaterialType[key] === materialtype;
+            if(validMaterial) break;
+        }
+        if(!validMaterial) {
+            throw new Error(`${materialtype} is not a valid material type!`);
+        }
+
+        this.cursor.materialtype = materialtype;
+        return this.cursor.materialtype;
+    }
```
samuel-zuk, line 703: If one doesn't exist there should be a "getMaterialType()" which should also be added to the reference, autocomplete, etc.

src/components/structural/MaterialType.js
```diff
@@ -0,0 +1,7 @@
+export const MaterialType = {
```
samuel-zuk, line 1: seeing as this isn't a component, it should probably be placed in "src/constants/MaterialType.js"

src/components/structural/View.js
beewyka819: Ah, I forgot to change the JSX in the `ent.tube` path. Going to push a fix for that.

The reason I can't just pass in ent.materialtype is that this doesn't work for some reason (can't access data.type, data.roughness, etc). In order to get it to work, I had to add `type:` to the schema and split up the data into its separate components in View.js. I never really figured out why ent.material doesn't have this problem, however I also can't find out where that component is registered, so I can't really compare the two.

src/myr/Myr.js
beewyka819: That was for an object that essentially acted as an enum for the different material types that was added before I realized I could just add autocompletion for the strings. Of course with the autocompleter for the strings this isn't really necessary, and in hindsight would really only make sense in a more strongly typed language, as in JS nothing would stop you from trying to read from a nonexistent member anyway. I'll remove it so that inputs for setMaterial() are more consistent and streamlined.

src/components/editor/customCompleter.js
```diff
@@ -205,6 +205,14 @@ export const customCompleter = {
             "yellowgreen"
         ];
 
+        let Materials = [
```
Jlu18, line 208: Minor change - change it to const

src/myr/reference.js
```diff
@@ -284,6 +284,24 @@ let transformations = [
...
+        parameters: [{type: "number", name: "roughness"}],
+        description: <span>The setRoughness function changes the roughness of the cursor within a range of 0 to 1. Custom roughness only applies to objects with MaterialType.PHYSICAL.</span>,
+        example: "setRoughness(0.5)"
+    },
+    {
+        name: "setMetalness",
+        parameters: [{type: "number", name: "metalness"}],
+        description: <span>The setMetalness function changes the metalness of the cursor within a range of 0 to 1. Custom metalness only applies to objects with MaterialType.PHYSICAL.</span>,
+        example: "setMetalness(0.5)"
+    },
```
Jlu18, line 304: "example" is used for a url link when user click on the example scene for the reference. So it should be just "setMaterialType" "setRoughness" and "setMetalness"

src/myr/reference.js
```diff
@@ -284,6 +284,24 @@ let transformations = [
         description: <span>The setTexture function changes the texture of the cursor. The texture's normal color is displayed when setTextureColoring() is false, otherwise the texture will be affected by the current color. Textures can be applied either by using the name of one of <button id="textures" onClick = {handleTextureOpen} style = {{backgroundColor:"white", color:"#47a0ff", padding:0, border:"none"}}>these textures</button> or by inserting a valid url (e.g. "bricks" or "https://learnmyr.org/img/MYR-Logo.png"). WidthRepeat and heightRepeat will change how many times the texture is displayed on the object in each direction. If widthRepeat and heightRepeat are not declared they will be set to default settings. Depending on the policy of the website they’re from, image URLs may not be usable, which will result in a blank Texture and Color. An empty setTexture() or setTexture("") statement will remove the current texture.</span>,
         example: "setTexture"
     },
+    {
+        name: "setMaterialType",
+        parameters: [{type: "MaterialType", name: "materialtype"}],
```
Jlu18, line 289: type should be string

src/utils/AFrameMaterials.js
```diff
@@ -0,0 +1,46 @@
+import AFRAME from "aframe";
+import {MaterialType} from "../components/structural/MaterialType.js";
+
+AFRAME.registerComponent("materialinfo", {
```
Jlu18, line 4: Can you add documentation for what this does?

src/tests/Myr.test.js
```diff
@@ -1,4 +1,5 @@
 import Myr from "../myr/Myr";
```
Jlu18, line 1: Please add tests for your newly added MYR APIs

Jlu18 (changes requested):

Jlu18:
Also, is there any way to see the list of the material available to the user?

Jlu18:
change setMaterialType to setMaterial

src/tests/Myr.test.js
samuel-zuk: (mentioned above) this should include getter functions

src/myr/Myr.js
```diff
@@ -670,7 +680,57 @@ class Myr {
...
+     * @param {number} metalness 
+     * @returns the metalness of the cursor
+     */
+    setMetalness = (metalness) => {
+        if(metalness < 0 || metalness > 1) {
+            throw new Error("Metalness must be between 0.0 and 1.0!");
+        }
+        this.cursor.metalness = metalness;
+        return this.cursor.metalness;
+    }
```
samuel-zuk, line 733: these functions need getters too-- getMetalness(), getRoughness()

src/myr/Myr.js
```diff
@@ -94,6 +98,9 @@ class Myr {
         this.cursor = {
             color: "red",
             texture: "",
+            materialType: MaterialType.SPECULAR,
+            roughness: 1.0,
```
samuel-zuk, line 102: super nitpicky but this will forever bug me-- is there a particular reason you put 0.0/1.0 here instead of 0/1?

samuel-zuk (changes requested):

src/myr/Myr.js
beewyka819: Static typing habits. Used to floats being separate from integers. Yay javascript numbers

kdvalin:
Closing this as this is now stale.

PR #501, opened 2021-07-14
Added documentation for missing functions and some clean up

Jlu18:
# The list of stuff I clean

collectionAction.js
-
- asyncCollections
  - rename the parameter id to uid to match with the rest
- deleteCollection
  - rename the parameter id to collectionID to match with the rest
- format some functions.

editorAction.js
-
- Deleted unused addPassword() function

projectAction.js
-
- asyncUserProj
  - rename the parameter id to uid
 
sceneAction.js
-
- I keep the unused functions (changeSettings, changeCamMode, changePerspective) because it might be useful but not sure.

component/collection/Collection.js
-
- Rename some of the functions to match with the collection

component/collection/SelectProject.js
-
- Remove the redundant code when checking for unsave code in handleChange()

component/courses/Course.js
-
- Remove the redundant code when checking for unsave code in nextLesson() and lastLesson()

component/courses/CourseSelect.js
-
- Deleted the constructor

customCompleter.js
-
- replace let with const for the keywords 
- rename keywords to MYRKeyWords to avoid confusion

AFramePhysics.js + View.js
-
- Removed the "basicmaterial" and "gridmaterial" component and replace with "shader:flat;" in View.js
- Removed the "outline" component

ProjectView.js + SceneConfigMenu.js
-
- Removed the unused password protection feature

ActionTypes.js 
-
- removed ADD_PW constant

Group.js
-
- Added check to see if the element is group in getEl()

Myr.js
-
- Removed duplicate makeUnPushable() and setCursorAttribute() function
- Removed addCModel and HALT
- Move functions that are not being used (Don't know whether should I delete it)
  - change, syncChange, sleep, infiniteLoopDetector, push, drop, and animate

myr/reference.js
-
- Reorder some of the references
- Delete the commented out code that are not needed

myr/textureReference.js & myr/modelReference.js
-
- Removed duplicate functions from reference.js
- Comment out example attribute since we don't have example scene for model or texture yet

reducers/scene.js
-
- reorder the cases

tests/myr.test.js
-
- removed addCModel from test suite

sceneAction.js, layouts/Collection.js, constants/ActionTypes.js, LayoutTypes.js, reducers/scene.js
-
- Rename to collection from classroom (old name for collection)

components/reference/ReferencePage.js
-
- Inherit ReferencePage from Reference so it reuse the same function

components/reference/TextureReference.js. ModelReference.js
-
- Removed the files because it wasn't used in anywhere else

ReferencePage.css
-
- Deleted TextureReferencePage.css and ModelReferencePage.css since it's same thing as ReferencePage.css and use that instead.

containers/ModelReference, containers/TextureReference.js, src/route.js
-
- Deleted ModelReference and TextureReference.js and connect the actions and props in Reference.js instead because the list of state and actions needed are same.


src/components/collection/SelectProject.js
```diff
@@ -23,21 +23,23 @@ const selectStyle = {
...
+
+    /**
+     * Fetch the selected scene when the project is selected from the collection
+     * @param {Event} projectID ID of the project to be fetch
+     */
     handleChange = (projectID) => {
         // show warning if there's unsaved change
         if(this.hasEditorChanged()){
-            if((window.confirm("Are you sure you want to continue?\nYou will lose any unsaved work!"))){
-                if (this.props.user && this.props.user.uid) {
```
kdvalin, line 34: So this code basically does nothing?  I'm guessing we're doing this since we're directing off to another page, which would refresh the editor anyways

src/components/structural/header/SceneConfigMenu.js
```diff
@@ -129,57 +128,47 @@ class ConfigModal extends Component {
         });
     };
 
-    handleClick = event => {
-        this.setState({ anchorEl: event.currentTarget, projectId: event.currentTarget.id });
-    };
+    // handleClick = event => {
```
kdvalin, line 131: Remove this instead of commenting it out

src/constants/ActionTypes.js
```diff
@@ -12,16 +12,17 @@ export const EDITOR_REFRESH = "EDITOR_REFRESH";
 export const EDITOR_RECOVER = "EDITOR_RECOVER";
 export const EDITOR_CHANGE = "EDITOR_CHANGE";
 export const EDITOR_UPDATE_SAVEDTEXT = "EDITOR_UPDATE_SAVEDTEXT";
-export const ADD_PW = "ADD_PW";
 export const SAVE_START = "SAVE_START";
+//unused constant
```
kdvalin, line 16: What's the unused constant?

src/constants/ActionTypes.js
```diff
@@ -12,16 +12,17 @@ export const EDITOR_REFRESH = "EDITOR_REFRESH";
...
 export const SAVE_FIN = "SAVE_FIN";
 export const REQ_RELOAD = "REQ_RELOAD";
 
-export const ASYNC_USER_PROJ = "ASYNC_USER_PROJ";
 export const SYNC_USER_PROJ = "SYNC_USER_PROJ";
-export const ASYNC_EXAMP_PROJ = "ASYNC_EXAMP_PROJ";
 export const SYNC_EXAMP_PROJ = "SYNC_EXAMP_PROJ";
 export const DELETE_PROJ = "DELETE_PROJ";
+//unused constant
```
kdvalin, line 23: Same here

src/myr/modelReference.js
```diff
@@ -1,169 +1,44 @@
...
-let model = [
+/**
+ * List of model reference
+ */
+const model = [
     {
         name: "sword",
         model: "sword",
-        example: "sword",
+        //example: "sword",
```
kdvalin, line 8: Remove the code, do not comment it out

src/myr/textureReference.js
```diff
@@ -1,249 +1,124 @@
...
+/**
+ * List of texture reference
+ */
+const texture = [
     {
         name: "bricks",
         image: "bricks",
         creator: "“Caveman Chuck” Coker",
-        example: "bricks",
+        //example: "bricks",
```
kdvalin, line 9: Again, remove, do not comment out

kdvalin (changes requested):

src/constants/ActionTypes.js
Jlu18: Those were the constants that weren't used it anywhere else. I thought there were some intention behind it so I didn't delete it but it's probably best to remove it.

src/components/collection/SelectProject.js
Jlu18: That part of the code is exactly same as the code below (when the code doesn't have any changes), it's redundant. So I flip the if statement to return if user doesn't want to proceed.

src/tests/Myr.test.js
```diff
@@ -752,13 +752,6 @@ describe("Component Animations", () => {
 });
 
 describe("Other Myr functionality", () => {
-    it("should add a model", () => {
-        myr.reset();
-        myr.assets = [];
-        myr.addCModel();
-        expect(myr.assets).toContainEqual({ id: "c-obj", src: "/img/c.obj" });
-    });
-
```
kdvalin, line 761: Is there a reason why this test is being dropped?

Jlu18: This is testing addCModel function from MYR API which create and it's deleted in MYR.js as well since it's not documented anywhere. There aremore APIs that are not documented anywhere but I wasn't sure whether if I should removed it so I moved it down in MYR.js

kdvalin (approved):

PR #511, opened 2021-08-11
Improve saving speed

Jlu18:
The majority of saving was coming from taking screenshot of scene and converting it to jpeg using built in aframe screenshot feature. By default, aframe capture screenshot with equirectangular (360 view of scene). And in order to have perspective image, aframe does a conversion form the equirectangular to perspective by modifying each pixels of image which is time consuming especially when the size of image is 2048 by 1024.

Since the getting screenshot was done synchronously, the page freezes until the conversion is finished which is why it takes time even though the saving itself is done asynchronously. 

I change the method of getting screenshot by creating a threejs renderer with preserveDrawingBuffer attribute to keep the buffer to convert to jpeg format. The screenshot still has to be done synchronously but the lag between saving scene and getting back to editor are significantly improved.

I also removed the spinner when the saving is occurred, and close the save tab immediately after the save button is press instead of after the saving is completed. Also render the scene after the saving is finished


Jlu18:
I found couple stuff that I want to fix with saving while testing. I will reopen it once I add the stuff

PR #534, opened 2022-01-10
Import/Export Feature

kdvalin:
This PR allows for scenes to be exported to a user's device and then imported at a later date.


Jlu18:
Can you update the backend [import/export PR](https://github.com/engaging-computing/MYR-backend/pull/160) so I can test it on my machine.

src/components/structural/header/ProjectView.js
```diff
@@ -423,6 +435,27 @@ class Project extends React.Component {
                                     }
                                     <span>&nbsp;</span>Preview
                                 </Button>
+                                <input
+                                    type="file"
+                                    accept="application/json"
+                                    ref={this.uploadRef}
+                                    onChange={(e) => this.props.importFunc(this.props.user.uid, e.target.files)}
```
Jlu18, line 442: Also returns error when user is not logged in.

src/components/structural/header/ProjectView.js
```diff
@@ -351,6 +356,13 @@ class Project extends React.Component {
                 </ListItemIcon>
                 <ListItemText inset primary="Send" />
             </MenuItem>
+            <MenuItem
+                onClick={() => this.props.exportFunc(this.props.user.uid, this.state.projectId)}>
+                <ListItemIcon>
+                    <Icon className="material-icons">download</Icon>
+                </ListItemIcon>
+                <ListItemText inset primary="Export Scene" />
+            </MenuItem>
```
Jlu18, line 365: Example project also has an option to export the scene but unable to.

src/components/structural/header/ProjectView.js
Jlu18: Probably hide the option if user is not logged in (or import the scene without saving as a scene)

src/components/structural/header/ProjectView.js
```diff
@@ -423,6 +435,27 @@ class Project extends React.Component {
                                     }
                                     <span>&nbsp;</span>Preview
                                 </Button>
+                                <input
```
Jlu18, line 438: Is there any reason why this input tag is outside of the import Button? It seems to be working fine but I think it should be a child of the Button.

Jlu18 (changes requested):
The import and export options should be disabled (or hide away) if the user is not logged in, as well as if the user is on example project tab (since it doesn't seem like you can export those)

src/components/structural/header/ProjectView.js
Jlu18: Exporting example scene still gave 404 error. Is this intentional or do we want to give user option to export the example scene?

Jlu18 (changes requested):

kdvalin:
Bumping to next version since I needed to update AFrame

Jlu18 (dismissed):

PR #532, opened 2022-01-07
Font Size

jf-kelly:
Added font size functionality for the editor. A dropdown menu button can be found below the editor next to the keyboard shortcuts button.

PR #537, opened 2022-02-07
Reference search

jf-kelly:
Created reference search bar, added functions and autocomplete feature, which required new node modules.


src/actions/userActions.js
```diff
@@ -0,0 +1,47 @@
+import * as types from "../constants/ActionTypes";
+
+const googleLoginRef = "apiv1/googlelogins";
+
+export function updateFontSize(fontSize){
+    return{ type: types.UPDATE_FONT_SIZE, fontSize };
+}
```
kdvalin, line 7: Relies on #532, in future base PRs on `dev` unless you need to.

Jlu18:
Looks like we accidentally based the branch out of #532 so we have lots of unnecessary file changes @jf-kelly, you should remake the PR with only necessary changes. Also this feature is going to beta not to dev

Jlu18:
> As for the failing CI, I've seen that before on #414. I needed to update aframe to fix that if memory serves. If this is happening regularly, I might need to move that back to the next release to unblock this issue.

@kdvalin The failed CI was the issue with the using the older version of the npm (version below 7) as it deletes the [hidden lockfiles](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#hidden-lockfiles) that's necessary for the newer version of the npm (which CI uses). So I added in setup wiki to check the npm version.

PR #554, opened 2022-03-25
Update nvmrc File to CI Version

kdvalin:
This updates the nvmrc file to use node v12.18 instead of v10.13.

This is a convenience update, so you can simply clone the repo and run `nvm install` to grab the correct node version.


bryanmontalvan (approved):
lgtm

PR #567, opened 2022-04-26
Add check to the asyncCollections to not process the data if api returns with 204

Jlu18:
This prevents the unhandled rejection error.


bryanmontalvan (approved):
lgtm, short and sweet :+1:

PR #576, opened 2022-06-02
increased visibility of buttons on the bottom of screen

r-maradiaga:
The two buttons for font size and keyboard shortcuts were hard to see. To change this, I changed the viewport height of the container that holds those buttons to be the same as the viewport height of the header.

### DIFF
--- a/src/actions/collectionActions.js
+++ b/src/actions/collectionActions.js
@@ -15,6 +15,9 @@ export function asyncCollections(uid) {
         if (uid) {
             let userCollections = [];
             fetch(collectRef, {headers: {"x-access-token": uid}}).then((data) => {
+                if(data.status === 204){
+                    return;
+                }
                 data.json().then((data) => {
                     data.forEach((doc) => {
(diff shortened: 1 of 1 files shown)


bryanmontalvan:
@RJMaradiaga Could you send me a screenshot of the new look?

RJMaradiaga:
image
image

bryanmontalvan:
@RJMaradiaga This is definitely an improvement, I think making both of the buttons(and their containers) a bit wider would be the best option. Could you mock it up?

RJMaradiaga:
I made two versions, both of which make the buttons taller and wider. 

This is one version that has the actual word "shortcuts" on the button. This option might be good because, when I just started on the website, I didn't have any idea of what the button did so I didn't even click it or hover over it. 

image

This version makes the buttons bigger and wider without the text. 

image

What do you think @bryanmontalvan ?

bryanmontalvan (dismissed):
Looks good to me :+1:

Issue #583, opened 2022-06-14
Reference examples not working


jf-kelly:
When you try to view any of the examples in the MYR reference tab, they do not show up. Instead, a blank grey page shows up.
To replicate:
1. Go to learnmyr.org
2. Click on the references button (? sign)
3. Click on any of the links for examples

PR #584, opened 2022-06-16
Fixed the bug with reference examples not showing up

jf-kelly:
This is a fix to issue #583. Originally, the reference examples were not working, but now they work. All I changed was where the editor is called in rendering the example, I added the property of user settings so that the font size would work.


bryanmontalvan (approved):
lgtm

PR #587, opened 2022-06-23
Improved visibility for font size and shortcuts buttons

RJMaradiaga:
Made the code editor slightly smaller to make more space for buttons. Additionally, increased the width and height of font size and shortcuts button.

Before: 
image

After: 
image


jdones01 (approved):
That looks a lot better! Awesome work!

PR #593, opened 2022-06-30
Fix deployment pipeline issues

kdvalin:
Updated instances of SSH fetching in package-lock.json to use HTTPS instead.

# Recommended Action Items

- [ ] Figure out why SSH fetching was occurring (look at dependabot first)
- [ ] Update git to only use HTTPS fetching in the pipeline (maybe something like [this](https://stackoverflow.com/questions/61051073/npm-install-from-git-use-https-instead-ssh))
- [ ] Add a container building step in PRs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8256,7 +8256,7 @@
             }
         },
         "node_modules/debug": {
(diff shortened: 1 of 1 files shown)


bryanmontalvan (approved):
lgtm will focus on the fixes

Issue #599, opened 2022-07-12
[BUG] Fix grid so that it takes up the entire scene floor


bryanmontalvan:
## Description
Currently the grid is not covering the entirety of the scene floor, this could've been caused from #578 (due to scaling issues)

## Preview
![image](https://user-images.githubusercontent.com/68972382/178583263-9b6d7863-0bc7-40c0-af83-63836f594916.png)

PR #605, opened 2022-07-14
Reverted grid back to original size per Bryan's request

jf-kelly:
The grid now does not line up with objects, but it fits the floor better than the other version


jf-kelly:
This is a fix for issue #599

bryanmontalvan (approved):
lgtm

Issue #602, opened 2022-07-12
Let's make the torus object an actual torus


bryanmontalvan:
## Description
Currently our `torus()` object is actually not a torus but some sort of "dumpling" looking shape. Let's make the torus an actual torus :)

As to what exact shape to use and where to get started, take a look at A-Frame's geomtry docs(https://aframe.io/docs/1.3.0/components/geometry.html)
## Preview
![image](https://user-images.githubusercontent.com/68972382/178586622-ab6eedde-04e6-4c59-9a25-54ab28c73879.png)

PR #607, opened 2022-07-14
Changed the torus shape to actually look like one

jf-kelly:
I changed the tubular radius of the torus shape so that it looks like what a torus should be (more like a donut). This is a fix for issue #602


bryanmontalvan (approved):
Awesome work @jf-kelly 
LGTM

PR #609, opened 2022-07-19
Expanded on replacing deprecated no-gutters

jf-kelly:
## Description

This PR is similar to Bryan's recent PR #608, as he was replacing the deprecated no-gutters with g-0, and I just did that in all the places that no-gutters is found. This is because I realized that the examples pages still had the larger middle bar, so every instance of no-gutters had to be changed.


bryanmontalvan (approved):
LGTM

PR #610, opened 2022-07-19
Scene title and description editing

jf-kelly:
## Description

This pull request is for a new feature: being able to edit the title and description of a scene that has already been saved. I added a new button in the open projects folder that when clicked, pulls up the option to change the title and description of the scene. However, the project must be open for this to work, so if the project is not open and the user presses the button, an error message pops up telling them to open the project first.

I can do a demo of this feature in the next meeting, as pictures are not great for explaining how it works.
## Preview

Preview of the button:

Preview of the error message:


bryanmontalvan (approved):
LGTM

I think the built in browser modal (`alert()`) works just fine on our use-case. Unless we're looking for a consistent UI design I think this will work for now

Great work!

Issue #598, opened 2022-07-12
[BUG] Keyboard-shortcut not being highlighted during tour


bryanmontalvan:
## Description
Currently when you **take the tour** and you get to step 12 you'll find that the keyboard-shortcut button is not highlighted. We want the button to be highlighted the same way every other component is highlighted in every other step

## Preview
![image](https://user-images.githubusercontent.com/68972382/178582651-a633e524-810e-4978-b59c-203db5f03e14.png)

PR #616, opened 2022-07-28
Tour Fix

jf-kelly:
## Description
The tour now shows both the keyboard shortcut button and the font size button. Previously, it only showed the font size button. 
## Related Issue
#598


RJMaradiaga:
@jf-kelly This change looks good. 

I know this isn't part of this PR but I thought it would be helpful to add it here. When I take the tour on dev, the 5th item, the explanation of the save button, highlights only half of the button. I'm on a windows laptop using chrome so I'm not sure if it's a problem for everybody but It's something worth checking to see if others have the same problem.

RJMaradiaga (approved):

Issue #614, opened 2022-07-28
[BUG] bad interaction between prior setScale when a group is created


fgmart:
The following code demonstrates that when a group is used and an object is added to it and draw, the prior scale parameters interact with scaling of the item drawn in the group:

```
setScale(0.25, 1, 0.25);
let g = group();
setPosition(0, 2, 0);  // works
setScale(2, 2, 2);     // prior scale combined with new scaling
setColor("blue");      // works
g.add(sphere());       // sphere drawn with combination (product?) of prior scaling and setScale immediately above
```
If you do a `resetCursor()` prior to creating the group, things will be OK (probably because scaling is reset to 1, 1, 1)

See https://learnmyr.org/scene/62e299927c990f00120740ec for a demo.

This should be fixed.

PR #617, opened 2022-08-04
Fix for scaling in groups

jf-kelly:
## Description
This fixes the issue where a previous setScale would influence a group that was created. If a setScale was called, then a group, then another setScale, the group would use the first setScale. Now, I have the group function call the resetCursor function so that it automatically resets the cursor and you won't have this bad interaction anymore.
## Preview
Before the fix:

After the fix:

## Related Issue
#614

### DIFF
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -2031,6 +2031,7 @@ class Myr {
      * Return a Entity that can be used to group elements together
      */
     group = () => {
+        this.resetCursor();
         let base = {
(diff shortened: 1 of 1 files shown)


RJMaradiaga:
I think that's a clever fix. Looks good!

RJMaradiaga (approved):

bryanmontalvan (approved):
lgtm,

Simple yet effective

Issue #613, opened 2022-07-21
[UI/UX] Remove underline from scene-title and add a hover-effect to the card


bryanmontalvan:
## Description

Currently in the `Open Project` menu we have scene title, scene preview, scene description, and scene options. We currently have the functionality in which we have two elements/components which `on-click` will redirect you to the chosen scene. Those two being the `scene-title` and `scene-preview`

I think it would be good to remove the on-hover underline effect from the scene-title and instead create a hover effect which shows an affect/animation on the entire scene card. 

## Preview
(How our current projects menu look)
<img width="1199" alt="image" src="https://user-images.githubusercontent.com/68972382/180284483-4b6d0303-b83d-4808-8365-9a95404c0302.png">

PR #619, opened 2022-08-04
Added new hover effect to project view

jf-kelly:
## Description
This pull request adds a new hover effect to the project view. Before, when a user hovered their mouse above a scene in project view, it would just underline the name of the scene. Now, when you hover above a scene, it no longer underlines the name, but instead makes the background of the entire scene card grey, to emphasize that you are opening that scene. 

Also, while making these changes, I found how to remove the underlines for all the links in the footer, which was another issue in dev, so now all that needs to be fixed in the footer is the spacing.

As for the hover effect, there are more things I can do with it, such as making the card slightly bigger, but for now, I am submitting the most simple effect, and we can discuss what other animations we might want for the hover effect.
## Related Issue
#613

### DIFF
--- a/src/components/structural/Footer.js
+++ b/src/components/structural/Footer.js
@@ -12,36 +12,36 @@ class Footer extends Component {
                         <span><strong>MYR:</strong></span>
                         <ul className="pl-2 list-inline">
                             <li className="pl-2 list-inline-item">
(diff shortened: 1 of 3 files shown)


bryanmontalvan (approved):
lgtm,

Nothing too flashy and that's exactly what we want. Appreciate you working on this!

Issue #611, opened 2022-07-20
[UI/UX] Footer-items are mis-aligned 


bryanmontalvan:
## Description

Currently on our dev branch the footer items alignment seems to be off (this seems to be an issue with padding or margin from bootstrap's `grid` class).

We recently updated bootstrap's version so this could possibly the reason why the footer items alignment changed

## Preview

**Production**
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/68972382/180019527-18a77134-ae42-4f7b-8fd1-5d128e8b9219.png">

**Dev**
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/68972382/180019476-71b34a06-50bf-4647-af08-acc47c690c4c.png">

PR #620, opened 2022-08-04
Items in the footer are aligned better than before

jf-kelly:
## Description
This PR fixes the issue with the items in the footer not being aligned properly. I simply removed the 'list-inline' property in the Footer.js file, and the items are more spread out.

This fix does not look exactly like it did before, but it does spread the items out a bit and looks better than what we currently have in dev.
## Related Issue
#611

### DIFF
--- a/src/components/structural/Footer.js
+++ b/src/components/structural/Footer.js
@@ -10,7 +10,7 @@ class Footer extends Component {
                 <div className="row m-3 text-center" >
                     <div className="col-lg-4 d-flex">
                         <span><strong>MYR:</strong></span>
-                        <ul className="pl-2 list-inline">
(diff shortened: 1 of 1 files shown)


RJMaradiaga:
The left does look better than what's in dev, but is there a way to make the copyright and privacy policy + terms of service buttons more similar to production? In production, it looks like the privacy policy and terms of service buttons are aligned left while the copyright is aligned at the center.

jf-kelly:
I added another commit that makes everything aligned a little better once again. The Bootstrap update changed certain things, like 'left' and 'right' are now 'start' and 'end', so it would possibly be worth looking into these changes to see if more has broken than we expected.

Issue #601, opened 2022-07-12
Make the speed-slider metrics make more sense


bryanmontalvan:
## Description
Currently our speed-slider goes from 0-1000. These numbers don't make much sense from a user's perspective, and as for another consideration is to not have 0 be an option in this slider. Because we don't want users not being able to move (they can stand still by not moving any movement keys).

To add onto this it would be a good idea to add labels to each end (left-side saying "slower" and right saying "faster"). Point is we want these selection of speed to be more intuitive. Also changing the range from 0-1000 to 0-10 would make more sense and be less obscure.

PR #621, opened 2022-08-04
Updated the Speed Slider UI to make more user-friendly

RJMaradiaga:
## Description
On the production of MYR, the speed slider has the value of speed when you hover over it. It started at 0, which seems like it would prevent the user from moving but instead the user could still move, just at the lowest speed. This pull request makes the speed slider in scene settings make more sense to the user. To do this, the arbitrary units of speed (which previously showed the user 0-1000) have been taken away and instead replacing them with words that show that moving the slider to the left makes the user slower and moving it to the right goes faster. 

## Preview
Before (Production version of MYR):
image

After: 
image

## Related Issue
Issue #601

### DIFF
--- a/src/components/structural/header/SceneConfigMenu.js
+++ b/src/components/structural/header/SceneConfigMenu.js
@@ -678,16 +678,19 @@ class ConfigModal extends Component {
                                                 ?
                                                 <div id="speed-config" className="col-12 pt-4">
                                                     <div className="row">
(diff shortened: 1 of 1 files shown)


jf-kelly (commented):
Looks good to me, I like the simplicity

jf-kelly (approved):
I like these changes, forgot to approve them in my last review

Issue #517, opened 2021-10-21
`shrink()` reference is unclear


kdvalin:
**Bug Description:**
The animation function `shrink` does nothing.

**Steps To Reproduce:**
Copy and paste the following code into [learnmyr.org](https://learnmyr.org)
```js
shrink(box());
```

**Expected behavior:**
The box to shrink and grow as the documentation for the animation outlines

**URL(s):**
An example of this bug can be seen [at this scene](https://learnmyr.org/scene/61719bddaf826a0013f5fd2e)

# Platforms the bug occurs on
## Desktop
 - OS: Debian 10
 - Browser: Firefox
 - Version: 92.0.1

Jlu18:
The shrink animation doesn't animate because the magnitude of the animation cursor is 1 by default and the animation is animating the scale current current scale to a value that divide scale by magnitude. Any scale divide by magnitude of 1 is scale, which means scale is changing to same value, so it doesn't have any visual effect.

If you set the magnitude to any value other than 1, you will see the effect.
This also applies to grow.

myr.js, line 1709 
` to: ${el.scale.x / magnitude} ${el.scale.y / magnitude} ${el.scale.z / magnitude};`

PR #622, opened 2022-08-09
Updated `shrink()` and `grow()` documentation to reference magnitude function

RJMaradiaga:
## Description
When looking at the reference page, some of the animation functions seem unusable without the use of other functions like `setMagnitude()`.The purpose of this PR is to make the `shrink()` and `grow()` functions easier to understand for any new user by referencing `setMagnitude()` in the description. Otherwise, the only way a user would know to use `setMagnitude()` would be to click the examples page. 

## Preview
Before: 
image
After:
image

## Related Issue
Issue #517

### DIFF
--- a/src/myr/reference.js
+++ b/src/myr/reference.js
@@ -625,13 +625,13 @@ const animations = [
     {
         name: "grow",
         parameters: [{ type: "string", name: "elementID" }],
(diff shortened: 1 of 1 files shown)


jf-kelly (commented):
It's good to reference the `setMagnitude()` function, the only change I would make is including the fact about setting it to a number other than 1 in the description for `grow()` as well

jf-kelly (approved):

Issue #502, opened 2021-07-14
spin() to accept multiple axis


Jlu18:
Current when you call spin(), it only rotates the object in y axis. I want it to accept in all axes.

I think we can achieve this by either create new functions call spinX, spinY, and spinZ, or set axes of rotation before calling spin() function.

PR #626, opened 2022-08-16
Update spin function

RJMaradiaga:
## Description
I created three new spin functions called spinX, spinY, and spinZ. These functions allow for greater modularity than the original spin function because now users have the option of spinning across all axes as opposed to just one axis. Additionally, these functions have a second optional parameter that specifies the spin type of the given object. These include "normal," which spins the object counterclockwise, "reverse," which spins the object clockwise, and "alternate," which spins the object both ways after a certain duration. 

The previous spin function has not yet been removed because it may cause issues in courses and example scenes. 

## Preview
New functions in the reference page: 
image

## Related Issue
Issue #502

### DIFF
--- a/src/myr/Myr.js
+++ b/src/myr/Myr.js
@@ -1410,6 +1410,85 @@ class Myr {
         return outerElId;
     };
 
+
+    /**
+     * Apply a spin animation on the X axis to the Aframe element which is passed as arg
+     * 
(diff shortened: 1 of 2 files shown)


jf-kelly (approved):
The changes look good to me, now the only thing, like you mentioned, is deciding what to do with the original `spin()` function. Other than that, these are positive changes!

Issue #600, opened 2022-07-12
[BUG] Default player speed is accelerated 


bryanmontalvan:
## Description
Currently in `dev` the player speed is much faster than it is on production (before the last batch of PRs merged into dev). The default speed is accelerated and the speed-slider isn't holding true to its given value. 

Ideas:
- This could also be caused by the change of scale in issue #578

PR #627, opened 2022-08-18
Fixed dev camera speed issues

jf-kelly:
## Description
Currently on dev, the camera speed is way too fast, most likely due to updating the a-frame version. However, despite looking through the a-frame docs and version changes, I was unable to find anything that would affect this. Ultimately, this PR simply adds a new variable that scales down the speed input from the slider, which seems to do the trick. The speeds closely line up with those in production.
## Preview
A screenshot is not applicable here, but I can do a demo if needed.

## Related Issue
#600

### DIFF
--- a/src/components/structural/View.js
+++ b/src/components/structural/View.js
@@ -292,6 +292,7 @@ class View extends Component {
      * @returns {HTMLElement} A-Frame camera elements with basic movement
      */
     basicMoveCam = () => {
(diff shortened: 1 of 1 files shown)


bryanmontalvan (approved):
It would be great to figure out what caused the movement speed to mess up, but in the mean time this a fix we should take!

Issue #346, opened 2019-08-09
Create user examples tab


jasondkiesling:
Create new tab on projects modal for user created scenes. This tab would allow users to submit a scene idea via this [form](https://forms.gle/PraaiKGm3eKUtZCy8). The MYR leadership team will then add scenes to this tab as they see fit.

PR #628, opened 2022-08-23
Added form to submit user example scenes

jf-kelly:
## Description
I was working on the idea of adding a new tab for user example projects in the "open project" modal, but I thought it would be better if user examples were also displayed in the "example scenes" tab. I believe it would be too confusing to have two tabs that are both showing example scenes. So, I just added text at the top of this already existing tab that prompts users to submit their scenes, and provides a link to the submission form.
As this is different from the original idea, it is still open to discussion of course.
## Related Issue
#346


bryanmontalvan (approved):
This is looking great @jf-kelly I was wondering how this looks on smaller devices(mobile specifically)? Apart from that it LGTM

PR #659, opened 2023-01-31
References fix

jf-kelly:
## Description
Fixes the issue with reference examples not showing up - this was fixed before but somehow broke again, so I just implemented the same changes again


bryanmontalvan (approved):
LGTM

PR #661, opened 2023-02-07
Reset button

StevenH1234:
## Description
 PR moves the reset button from the scene configuration drop-down menu to the header to make it easier for the user to reset their scene 

## Preview
attached. 
Screen Shot 2023-02-07 at 4 28 48 PM


jf-kelly (changes requested):
Looks good Steven, I like the change. Would you also be able to remove the original button from within the settings tab?

jf-kelly (approved):
Looks good, thanks Steven!

Issue #657, opened 2023-01-26
[BUG] Vertical Movement is disabled for users on a desktop


rjmara:
**Bug Description:**
On the desktop, the user is limited to horizontal movement. User's can still look around in any direction, but they cannot use the space or ctrl button to move up or down. This happened after the servers restarted and is isolated to the desktop version of the website.

**Additional context:**
If a newer version of MYR cannot be restored, the problem could be found directly under `src->components->structural->view` in the basicMoveCam function. The camera of mobile MYR has a movement-controls class which enables fly. The same is not true for the desktop version. Additionally, there is a file called `WASDPlusControls.js` that creates an aframe component which includes WASD, space, and ctrl buttons for movement.

PR #662, opened 2023-02-14
Fixed Vertical Movement

jf-kelly:
## Description
On both production and dev, vertical movement had stopped working. I changed the View.js file, because in some previous commit, `wasd-plus-controls` was changed to `wasd-controls`, which only controls horizontal movement. So, the `a-camera` entity now has both of these attributes, so that, if needed, we can change the way vertical and horizontal speeds are handled independently of each other.

## Related Issue
#657

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8277,7 +8277,7 @@
             }
         },
         "node_modules/debug": {
-            "resolved": "https://github.com/ngokevin/debug.git#ef5f8e66d49ce8bc64c6f282c15f8b7164409e3a"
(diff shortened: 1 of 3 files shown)


bryanmontalvan (approved):
Great catch with this John!

My only complaint is that you included the change of capitalization of `userSettings` in the `layouts/ReferenceExample.js` file. Personally I think PRs should be isolated for specific issues. An example being yourself or someone could go through and correct the captializations of every `userSettings` declaration. 

However, that's more of a quirk than anything else. Apart from that everything looks great!

PR #671, opened 2023-04-06
Downgraded A-Frame version to fix VR bug

banstack:
## Description

This PR downgrades aframe from 1.3 to 1.2, in turn fixing the VR not working on mobile


jf-kelly (approved):
Looks good to me, nice fix

jf-kelly (commented):
I think a possible issue with this PR failing in dev is on certain lines, the url is changed from https to git+ssh. The error that is thrown has to do with ssh, so this would be my best guess as to why it fails the ci checks

PR #674, opened 2023-05-15
Updating styles for the footer and course select

r-maradiaga:
## Description
When we updated bootstrap, the elements in the footer became misaligned. To fix this, I used the classes from the current 5.1 version of bootstrap. Additionally, when selecting courses, the current version of MYR makes the header and paragraph of each course underlined. I just changed the style for this so that each part of the card does not have an underline upon opening the course menu. Then, when someone hovers over the course, only the header becomes underlined.


jf-kelly (approved):
Looks good

